PR #1, opened 2024-09-05
Add `__slots__` to classes for performance improvements

harshil21:
See https://stackoverflow.com/questions/472000/usage-of-slots for benefits of slots. 

Additionally, I ran the code through a code formatter, [_Ruff_](https://github.com/astral-sh/ruff). I'll also create a separate PR introducing some lint rules (reduces chances of bugs)

Code is UNTESTED.

### DIFF
--- a/Scripts/test_logger.py
+++ b/Scripts/test_logger.py
@@ -3,7 +3,7 @@ from Airbrakes.imu import IMUDataPacket
 from Airbrakes.logger import Logger
 
 
-CSV_HEADERS = ["state", "extension"] + list(IMUDataPacket(0.0).__dict__.keys())
+CSV_HEADERS = ["state", "extension"] + list(IMUDataPacket(0.0).__slots__)
 
 
 def main():
--- a/Airbrakes/airbrakes.py
+++ b/Airbrakes/airbrakes.py
@@ -12,6 +12,16 @@ class AirbrakesContext:
     Read more about the state machine pattern here: https://www.tutorialspoint.com/design_pattern/state_pattern.htm
     """
 
+    __slots__ = (
+        "logger",
+        "servo",
+        "imu",
+        "state",
+        "shutdown_requested",
+        "current_extension",
+        "current_imu_data",
+    )
+
     def __init__(self, logger: Logger, servo: Servo, imu: IMU):
         self.logger = logger
         self.servo = servo
--- a/Airbrakes/imu.py
+++ b/Airbrakes/imu.py
@@ -8,15 +8,25 @@ class IMUDataPacket:
     roll of the rocket and the timestamp of the data.
     """
 
+    __slots__ = (
+        "timestamp",
+        "acceleration",
+        "velocity",
+        "altitude",
+        "yaw",
+        "pitch",
+        "roll",
+    )
+
     def __init__(
-            self,
-            timestamp: float,
-            acceleration: float = None,
-            velocity: float = None,
-            altitude: float = None,
-            yaw: float = None,
-            pitch: float = None,
-            roll: float = None
+        self,
+        timestamp: float,
+        acceleration: float = None,
+        velocity: float = None,
+        altitude: float = None,
+        yaw: float = None,
+        pitch: float = None,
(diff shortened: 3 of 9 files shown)


Scripts/test_logger.py
```diff
@@ -3,7 +3,7 @@
 from Airbrakes.logger import Logger
 
 
-CSV_HEADERS = ["state", "extension"] + list(IMUDataPacket(0.0).__dict__.keys())
+CSV_HEADERS = ["state", "extension"] + list(IMUDataPacket(0.0).__slots__)
```
JacksonElia, line 6: this is nice

JacksonElia (commented):
My understanding of slots was that they could help improve performance when many instances of a class are made, but when there's just one, I thought it didn't really matter.

Do you think it is necessary for the IMU, Servo, Logger, and AirbrakesContext? I would argue that it adds complexity for little/no performance gain.

That being said, it is definitely a great idea for the IMUDataPacket class. Those are created every loop, so I think it would help performance for sure.

Also, I like using a linter, I was meaning to get us on one.

JacksonElia:
> My understanding of slots was that they could help improve performance when many instances of a class are made, but when there's just one, I thought it didn't really matter.
> 
> Do you think it is necessary for the IMU, Servo, Logger, and AirbrakesContext? I would argue that it adds complexity for little/no performance gain.
> 
> That being said, it is definitely a great idea for the IMUDataPacket class. Those are created every loop, so I think it would help performance for sure.
> 
> Also, I like using a linter, I was meaning to get us on one.

Okay I read more into the thread you posted, and this was what they I said:
```
Do not "only use __slots__ when instantiating lots of objects"
I quote:

"You would want to use __slots__ if you are going to instantiate a lot (hundreds, thousands) of objects of the same class."

Abstract Base Classes, for example, from the collections module, are not instantiated, yet __slots__ are declared for them.

Why?

If a user wishes to deny __dict__ or __weakref__ creation, those things must not be available in the parent classes.

__slots__ contributes to reusability when creating interfaces or mixins.

It is true that many Python users aren't writing for reusability, but when you are, having the option to deny unnecessary space usage is valuable.
```

I don't really find these reasons compelling enough. I think it makes the code a little less readable (especially for those unfamiliar with __slots__) and I don't think saving minute amounts of space really matters here. Additionally, the [intended use](https://wiki.python.org/moin/UsingSlots) of __slots__ seems to be when you have a lot of objects, so idk if I like going against that.

All of that being said, I'm curious what you think. Did I misinterpret any of this? Also, if you think it would produce a measurable  performance increase then I'm all for it.

harshil21:
> Do you think it is necessary for the IMU, Servo, Logger, and AirbrakesContext? I would argue that it adds complexity for little/no performance gain.

imo there is basically no added complexity when slots are added for a small codebase like ours. Personally, when I [implemented slots](https://github.com/python-telegram-bot/python-telegram-bot/pull/2345) for a big library (~50k+ LoC), while it did initially require some refactoring, but in the end it actually helped promote good coding practices, such as not adding attributes (accidently) to classes, e.g. `MyClass().some_random_attr = 1`, and users [reported](https://t.me/pythontelegrambotgroup/599826) tremendous memory savings and slight performance increases, despite not having "thousands" of classes being initialized.

> Abstract Base Classes, for example, from the collections module, are not instantiated, yet __slots__ are declared for them.

I would also like to add that most built in classes in the standard library now have `__slots__` defined for them, and it's really an easy optimization trick.

> I think it makes the code a little less readable (especially for those unfamiliar with slots) 

While I would agree that it's not understood at first glance, it's not really a reason to ignore a feature of the language. A short comment about slots linking to any of the articles would suffice imo.

Additionally, we _could_ "automate" the creation of slots through [dataclasses](https://docs.python.org/3.12/library/dataclasses.html#dataclasses.dataclass), but that is actually some added complexity which I think we should avoid, since it doesn't actually provide any further benefits, other than less lines of code.


The little performance gain for our use case is comparable to upgrading from py 3.9 to py 3.12 (without build time optimizations) 😉

PR #2, opened 2024-09-08
Add Ruff as a formatter and linter

harshil21:
Adds Ruff as a formatter and linter. 

The list of rules enabled can be found in `pyproject.toml`. See https://docs.astral.sh/ruff/rules/ if you want to know what each rule does, and if you have any suggestions on more rules we could add.

Also adds an integration with Github Actions, so every PR will have to meet the quality standards.

I would also recommend switching max line length to 100 (the default which I've seen most codebases use), [instead of PEP-8's recommended line length of 79](https://peps.python.org/pep-0008/#maximum-line-length).

Additionally, I changed a little bit of the logging `__init__` logic, there is now no while loop, and we directly find which number to use for the log file.

### DIFF
--- a/Airbrakes/imu.py
+++ b/Airbrakes/imu.py
@@ -1,4 +1,7 @@
+"""Module for interacting with the IMU (Inertial measurement unit) on the rocket."""
+
 import multiprocessing
+
 import mscl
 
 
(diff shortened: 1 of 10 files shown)


Airbrakes/logger.py
```diff
@@ -14,21 +17,22 @@ class Logger:
...
 
     def __init__(self, csv_headers: list[str]):
         self.csv_headers = csv_headers
 
-        self.log_path = os.path.join("logs", "log1.csv")
+        log_dir = Path("logs")
+        log_dir.mkdir(parents=True, exist_ok=True)
+
+        # Get all existing log files and find the highest suffix number
+        existing_logs = list(log_dir.glob("log_*.csv"))
```
harshil21, line 29: Note the new file name scheme I'm using here, e.g. `log_1.csv` instead of `log1.csv`.

JacksonElia: cleaner than what i had

JacksonElia (approved):

Airbrakes/imu.py
```diff
@@ -74,11 +74,12 @@ def __init__(self, port: str, frequency: int, upside_down: bool):
...
         # Starts the process that fetches data from the IMU
-        self.data_fetch_process = multiprocessing.Process(target=self._fetch_data_loop, args=(port, frequency,
-                                                                                              upside_down))
+        self.data_fetch_process = multiprocessing.Process(
+            target=self._fetch_data_loop, args=(port, frequency, upside_down)
+        )
         self.data_fetch_process.start()
 
-    def _fetch_data_loop(self, port: str, frequency: int, upside_down: bool):
+    def _fetch_data_loop(self, port: str, frequency: int, _: bool):
```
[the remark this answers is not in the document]
harshil21: no, but it did tell me that variable was unused

PR #4, opened 2024-09-11
Architecture and logging improvements, packets batch processing

harshil21:
- Fixes a race condition where `get_imu_data` might have accessed a parameter which may not be in sync with other parameters for that particular instant (the loop continously updates values, and we fetched data while that was running) -> Fixed by using classes.

- Introduces batch processing of `IMUDataPackets` in `AirbrakesContext.update`. Ready packets are fetched in the `.update()` loop and it is processed in the new `IMUDataProcessor` class which calculates moving averages, max altitude, etc.

- Logging is now done by the `csv` module instead of `logging`. It's a cleaner interface and reduces manual updating of fields.

- `IMUDataPacket` now split into two - `RawDataPacket` and `EstimatedDataPacket`

- constants file added, package renaming, and better organization

- better type hinting and more documentation in general

Most code is **UNTESTED**.


Airbrakes/imu.py
```diff
@@ -83,17 +83,28 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
         """
         This is the loop that fetches data from the IMU. It runs in parallel with the main loop.
         """
-        while self.running.value:
-            # Connect to the IMU
-            connection = mscl.Connection.Serial(port)
-            node = mscl.InertialNode(connection)
```
harshil21, line 89: i'm not sure why we keep reconnecting to the imu every loop. Once should be enough isn't it?

Aman-Anas: that should be in init yeah, no need to keep making serial objects

JacksonElia: sorry I must have put it in the wrong spot. It shouldn't be in init, it should be in the process but above the while loop

Airbrakes/airbrakes.py
```diff
@@ -38,19 +38,29 @@ def __init__(self, logger: Logger, servo: Servo, imu: IMU):
...
         # Gets the current extension and IMU data, the states will use these values
         self.current_extension = self.servo.current_extension
-        self.current_imu_data = self.imu.get_imu_data()
+        data_packets = []
+
+        # Let's get 50 data packets to ensure we have enough data to work with.
+        # 50 is an arbitrary number for now - if the time resolution between each data packet is
+        # 2ms, then we have 2*50 = 100ms of data to work with at once.
+        while len(data_packets) < 50:
+            # get_imu_data() gets the "first" item in the queue, i.e, it *may* not be the most
```
JacksonElia, line 53: Queues are FIFO so it will really be the oldest data that you get first, but its fine bc we have a time stamp

Airbrakes/airbrakes.py
```diff
@@ -38,19 +38,29 @@ def __init__(self, logger: Logger, servo: Servo, imu: IMU):
...
         """
         # Gets the current extension and IMU data, the states will use these values
         self.current_extension = self.servo.current_extension
-        self.current_imu_data = self.imu.get_imu_data()
+        data_packets = []
+
+        # Let's get 50 data packets to ensure we have enough data to work with.
+        # 50 is an arbitrary number for now - if the time resolution between each data packet is
+        # 2ms, then we have 2*50 = 100ms of data to work with at once.
+        while len(data_packets) < 50:
```
JacksonElia, line 52: I don't know if I love this -- I don't like that how fast the system runs is linked with how fast/how much data it is given. This should work and *should* be fine, but I don't think its good practice. It also feels wrong having the context hold onto the data packets.

I think I would rather the imu class hold onto its most recent data packets. Maybe in this update we have a method first check if there is any data, then if there is, pop all of it off and log it, and then have the states use the most recent data point?

This is a loose idea and it very well might have a lot of flaws, so lmk what you think

Airbrakes/imu.py
```diff
@@ -83,17 +83,28 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
+        # Connect to the IMU
```
JacksonElia, line 86: good catch

JacksonElia (changes requested):

Airbrakes/airbrakes.py
harshil21: yeah, we _could_ process on the latest data, but I fear potentially having discarded data packets which might be critical to determining the state, and just logging in general.

Airbrakes/airbrakes.py
harshil21: > I don't know if I love this -- I don't like that how fast the system runs is linked with how fast/how much data it is given. This should work and _should_ be fine, but I don't think its good practice. 

Yeah, I did this because we need to process data over time, in real time. Technically, if the IMU is sending erroneous data or very delayed data, we are cooked and can't do anything anyway. 

I just got an idea on how to improve this - We could calculate the sampling rate (via e.g. moving average) of the IMU in real time, so if it drops below a certain threshold we can log that / error out / actually replace this "50" number and calculate a good length. That's of course a tiny performance overhead, but it should be worth the data and insight?

> It also feels wrong having the context hold onto the data packets. I think I would rather the imu class hold onto its most recent data packets

Agreed, that it should be abstracted away by the `IMU` class.

> and then have the states use the most recent data point?

but that's the thing, don't we want to calculate the state update, and that needs a couple of data packets? How can we determine the next state by looking at a single data point?

Airbrakes/airbrakes.py
```diff
@@ -38,19 +38,29 @@ def __init__(self, logger: Logger, servo: Servo, imu: IMU):
...
-        Called every loop iteration. Depending on the current state, it will do different things. It
-        is what controls the airbrakes and chooses when to move to the next state.
+        Called every loop iteration from the main process. Depending on the current state, it will
+        do different things. It is what controls the airbrakes and chooses when to move to the next
+        state.
         """
         # Gets the current extension and IMU data, the states will use these values
         self.current_extension = self.servo.current_extension
-        self.current_imu_data = self.imu.get_imu_data()
+        data_packets = []
```
harshil21, line 47: should be collections.deque for higher performance -> O(1) pop and append

Airbrakes/imu.py
```diff
@@ -104,38 +115,32 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
                             pass
 
-                # Update the timestamp after processing all data points
-                self.latest_data["timestamp"] = timestamp
+                # Put the latest data into the shared queue
+                self.data_queue.put(imu_data_packet)
 
     def get_imu_data(self) -> IMUDataPacket:
         """
         Gets the latest data from the IMU.
```
harshil21, line 135:
```suggestion
        Gets the last available data packet from the IMU.
```

Airbrakes/imu.py
```diff
@@ -104,38 +115,32 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
+
+        :return: an IMUDataPacket object containing the latest data from the IMU. If a value is
+        not available, it will be None.
         """
-        # Create an IMUDataPacket object using the latest data
-        return IMUDataPacket(
-            # When you use .get() on a dictionary, it will return None if the key doesn't exist
-            timestamp=self.latest_data.get("timestamp", 0.0),
-            acceleration=self.latest_data.get("acceleration"),
-            velocity=self.latest_data.get("velocity"),
```
harshil21, line 133: we are currently missing this

Airbrakes/airbrakes.py
```diff
@@ -1,10 +1,15 @@
 """Module which provides a high level interface to the air brakes system on the rocket."""
 
-from Airbrakes.imu import IMU, IMUDataPacket
+from typing import TYPE_CHECKING
+
+from Airbrakes.imu import IMU, IMUDataPacket, RollingAverages
```
Aman-Anas, line 5: if everything is going to be in an `airbrakes` package you might as well use relative imports `from .imu import IMU, bla`

that's typically the better way IIRC

Aman-Anas: also it doesn't really matter but package names should be lowercase so airbrakes not `Airbrakes`

harshil21: absolute imports are recommended per [PEP-8](https://peps.python.org/pep-0008/#imports), its way more clear

harshil21: agreed to make it lowercase tho

harshil21:
an approving review + merge would be nice, this PR is getting too large and out of scope.

airbrakes/imu/imu.py
```diff
@@ -0,0 +1,136 @@
...
+                                # These specific data points are matrix's rather than doubles
+                                case "estAttitudeUncertQuaternion" | "estOrientQuaternion":
+                                    matrix = data_point.as_Matrix()
+                                    # Converts the [4x1] matrix to the X, Y, Z, and W of the quaternion
+                                    quaternion_tuple = tuple(matrix[i, 0] for i in range(matrix.rows()))
+                                    # Sets the X, Y, Z, and W of the quaternion to the data packet object
+                                    setattr(imu_data_packet, channel + "X", quaternion_tuple[0])
+                                    setattr(imu_data_packet, channel + "Y", quaternion_tuple[1])
+                                    setattr(imu_data_packet, channel + "Z", quaternion_tuple[2])
+                                    setattr(imu_data_packet, channel + "W", quaternion_tuple[3])
```
harshil21, line 95:
```suggestion
                                    setattr(imu_data_packet, f"{channel}X", quaternion_tuple[0])
                                    setattr(imu_data_packet, f"{channel}Y", quaternion_tuple[1])
                                    setattr(imu_data_packet, f"{channel}Z", quaternion_tuple[2])
                                    setattr(imu_data_packet, f"{channel}W", quaternion_tuple[3])
```
can you change to f-strings, they are faster

harshil21 (commented):
ruff is failing

airbrakes/imu/imu_data_packet.py
```diff
@@ -0,0 +1,152 @@
...
+        self.gpsCorrelTimestampWeekNum = gpsCorrelTimestampWeekNum
+        self.scaledAccelX = scaledAccelX
+        self.scaledAccelY = scaledAccelY
+        self.scaledAccelZ = scaledAccelZ
+        self.scaledGyroX = scaledGyroX
+        self.scaledGyroY = scaledGyroY
+        self.scaledGyroZ = scaledGyroZ
+
+
+class EstimatedDataPacket(IMUDataPacket):
```
Aman-Anas, line 77: Could maybe use `msgspec` Structs (for speed) or a `dataclass` with slots enabled (for readability)

Aman-Anas: https://jcristharif.com/msgspec/structs.html

harshil21: yeah i'm fine with dataclass with slots for those classes. we aren't doing json (de)serialization so I don't see the need for that

harshil21: ok i'm down to use msgspec, in a different PR tho

airbrakes/imu/imu.py
```diff
@@ -0,0 +1,136 @@
...
+                                    setattr(imu_data_packet, f"{channel}W", quaternion_tuple[3])
+                                case _:
+                                    # Because the attribute names in our data packet classes are the same as the channel
+                                    # names, we can just set the attribute to the value of the data point.
+                                    setattr(imu_data_packet, channel, data_point.as_float())
+
+                # Put the latest data into the shared queue
+                self.data_queue.put(imu_data_packet)
+
+    def get_imu_data_packet(self) -> IMUDataPacket | EstimatedDataPacket:
```
JacksonElia, line 104: Shouldn't this just return IMUDataPacket?

airbrakes/imu/imu.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Note: If `get_imu_data_packet` is called slower than the frequency set, the data will not
+        be the latest, but the first in the queue.
+
+        :return: an IMUDataPacket object containing the latest data from the IMU. If a value is
+            not available, it will be None.
+        """
+        return self.data_queue.get()
+
+    def get_imu_data_packets(self) -> collections.deque[IMUDataPacket | EstimatedDataPacket]:
```
JacksonElia, line 116: same for this one

airbrakes/imu/imu_data_packet.py
```diff
@@ -0,0 +1,152 @@
+"""Module for describing the datapackets from the IMU"""
+
+
+def _mro_slots(obj) -> list[str]:
```
JacksonElia, line 4: can you explain this method better? I'm a little lost

airbrakes/state.py
```diff
@@ -1,10 +1,17 @@
...
 
-from Airbrakes.airbrakes import AirbrakesContext
+from abc import ABC, abstractmethod
+from typing import override
 
+from airbrakes.airbrakes import AirbrakesContext
 
-class State:
+
+class State(ABC):
```
JacksonElia, line 9: I like this

JacksonElia (approved):

airbrakes/imu/imu.py
harshil21: oops yes.

airbrakes/imu/imu_data_packet.py
harshil21: Basically when you access `.__slots__` of an instance, it only gives you the slots defined for that class. But we also want the slots of its parent class, i.e. IMUDataPacket for the `timestamp`. MRO = method resolution order - https://docs.python.org/3/library/stdtypes.html#class.__mro__

JacksonElia: oh cool

PR #5, opened 2024-09-14
Better encapsulation for the `Logger` & `IMU` class and fix a bug

harshil21:
- Explicit stop/start methods for logger and imu, and make some members private
- Fix bug in log() method of Logger
- Prevent potential of of logging loop getting stuck when no items are queue and stop() is called

To be merged after #4

### DIFF
--- a/airbrakes/constants.py
+++ b/airbrakes/constants.py
@@ -46,4 +46,8 @@ CSV_HEADERS = [
     *list(IMUDataPacket.__slots__),
     *list(RawDataPacket.__slots__),
     *list(EstimatedDataPacket.__slots__),
-]  # TODO: this needs to be used in the logger module
+]
+
+# The signal to stop the logging process, this will be put in the queue to stop the process
+# see stop() and _logging_loop() for more details.
+STOP_SIGNAL = "STOP"
--- a/airbrakes/constants.py
+++ b/airbrakes/constants.py
@@ -1,6 +1,12 @@
 """Contains the constants used in the airbrakes module"""
 
+from pathlib import Path
+
 from airbrakes.imu.imu_data_packet import EstimatedDataPacket, IMUDataPacket, RawDataPacket
+
+# -------------------------------------------------------
+# Servo Configuration
+# -------------------------------------------------------
 
 # The pin that the servo's data wire is plugged into, in this case the GPIO 12 pin which is used for PWM
 SERVO_PIN = 12
@@ -9,13 +15,29 @@ SERVO_PIN = 12
 MIN_EXTENSION = -1
 MAX_EXTENSION = 1
 
-# Should be checked before launch
-UPSIDE_DOWN = True  # TODO: Currently not factored in the implementation
+# -------------------------------------------------------
+# IMU Configuration
+# -------------------------------------------------------
+
 # The port that the IMU is connected to
 PORT = "/dev/ttyACM0"
 
 # The frequency in Hz that the IMU will be polled at
 FREQUENCY = 100  # TODO: Remove this since we don't/can't control the frequency from the code.
+
+# -------------------------------------------------------
+# Orientation Configuration
+# -------------------------------------------------------
+
+# Should be checked before launch
+UPSIDE_DOWN = True  # TODO: Currently not factored in the implementation should be added to DataProcessor
+
+# -------------------------------------------------------
+# Logging Configuration
+# -------------------------------------------------------
+
+# The path of the folder to hold the log files in
+LOGS_PATH = Path("logs")
 
 # The headers for the CSV file
 CSV_HEADERS = [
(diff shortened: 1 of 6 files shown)


airbrakes/logger.py
```diff
@@ -35,14 +37,32 @@ def __init__(self):
...
             writer.writeheader()
 
-        # Makes a queue to store log messages, basically it's a process-safe list that you add to the back and pop from
-        # front, meaning that things will be logged in the order they were added
-        self.log_queue = multiprocessing.Queue()
-        self.running = multiprocessing.Value("b", True)  # Makes a boolean value that is shared between processes
+        # Makes a queue to store log messages, basically it's a process-safe list that you add to
+        # the back and pop from front, meaning that things will be logged in the order they were
+        # added
+        self._log_queue: multiprocessing.Queue[IMUDataPacket] = multiprocessing.Queue()
```
JacksonElia, line 43: Shouldn't this type be a Queue of dictionaries?

airbrakes/logger.py
```diff
@@ -35,14 +37,32 @@ def __init__(self):
...
+        self._running = multiprocessing.Value("b", False)  # Makes a boolean value that is shared between processes
 
         # Start the logging process
-        self.log_process = multiprocessing.Process(target=self._logging_loop)
-        self.log_process.start()
+        self._log_process = multiprocessing.Process(target=self._logging_loop)
+
+        # The signal to stop the logging process, this will be put in the queue to stop the process
+        # see stop() and _logging_loop() for more details.
+        self._stop_signal = "STOP"
```
JacksonElia, line 51: if we define the queue as a being a queue of dicts, then this has to be a dict I guess?

airbrakes/logger.py
harshil21: yes thanks should be updated

airbrakes/logger.py
```diff
@@ -51,13 +71,12 @@ def _logging_loop(self):
...
+            while self._running.value:
                 # Get a message from the queue (this will block until a message is available)
-                # Because there's no timeout, it will wait indefinitely until it gets a message
-                # -- this is fine in practice, as >100 messages a second should be added to the
-                # queue, but if for some reason the queue is empty, it will block forever and
-                # stop() won't work
-                message_fields = self.log_queue.get()
+                # Because there's no timeout, it will wait indefinitely until it gets a message.
+                message_fields = self._log_queue.get()
+                if message_fields == self._stop_signal:
```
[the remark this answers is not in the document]
harshil21: well yeah I suppose we might be able to

airbrakes/logger.py
harshil21: unfortunately it's a different process so we'd have to do make it shared. `object()` would be optimal but that didn't work, probably because of the same reason. Somehow defining it as a string works

JacksonElia: Wait I thought because we use .put, thats what shares it with the process via the queue? So a dict should be fine? If it has to stay as a string, I think we should just define _log_queue as a general queue rather than one for a specific type

harshil21: yeah it works to put it in the queue, but when we compare it in the `_logging_loop` (`.get() != self._stop_signal`) that time it doesn't work. But let me check that again

airbrakes/constants.py
```diff
@@ -1,27 +1,49 @@
...
+
 # The headers for the CSV file
 CSV_HEADERS = [
     "state",
     "extension",
     *list(IMUDataPacket.__slots__),
     *list(RawDataPacket.__slots__),
     *list(EstimatedDataPacket.__slots__),
-]
+]  # TODO: this needs to be used in the logger module
```
harshil21, line 49: it is being used?

airbrakes/logger.py
```diff
@@ -55,13 +55,6 @@ def start(self):
         """
         self._log_process.start()
 
-    @property
-    def is_running(self) -> bool:
```
harshil21: aw but I liked that

airbrakes/constants.py
JacksonElia: It is in ```writer = csv.DictWriter(file_writer, fieldnames=CSV_HEADERS)```

airbrakes/logger.py
JacksonElia: we weren't using it anywhere, do you think we should/we will use it?

airbrakes/constants.py
harshil21: yeah that's what it is being used, so why the comment

airbrakes/logger.py
harshil21: we probably won't, but it's a nice to have to know the logger is actually running.

JacksonElia (approved):

PR #9, opened 2024-09-15
Clean up/optimizations

JacksonElia:
It looks like hasattr and setattr are really slow, so I changed it to updating the dict directly. It might be slightly faster if we stored all of the attributes, and then initialized the struct with everything it needs at the beginning, but idk if that's necessary.
https://stackoverflow.com/questions/12798653/does-setattr-and-getattr-slow-down-the-speed-dramatically


airbrakes/imu/imu.py
```diff
@@ -92,24 +92,26 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
+                                f"{channel}Z": quaternion_tuple[2],
+                                f"{channel}W": quaternion_tuple[3],
+                            })
+                        else:
+                            # General case, add the attribute to the dictionary
+                            data_to_set[channel] = data_point.as_float()
+
+                        # Update the attributes of the imu_data_packet in one call, this is much faster than setting
+                        # each attribute individually
+                        imu_data_packet.__dict__.update(data_to_set)
```
harshil21, line 114: this simply does not work for `msgspec` structs and for classes with `__slots__`. I don't believe there is any alternative either. `setattr` is pythonic and the recommended way to set attributes dynamically.

harshil21 (changes requested):
The link to the stackoverflow is over 10 years old. Latest python versions are heavily optimized

I profiled it using `timeit`:

```python
setup = "from airbrakes.imu.imu_data_packet import RawDataPacket; obj=RawDataPacket(0, 2, 3, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6)"
result = timeit.repeat(stmt="setattr(obj, 'scaledGyroZ', 5)", setup=setup, repeat=5, number=100000)
print("Time taken by setattr: ", result)
```
```
Time taken by setattr:  [0.005478162987856194, 0.003392255021026358, 0.003619926981627941, 0.0033833660127129406, 0.003353002975927666]
``` 

And I also profiled setattr vs dict update performance in general:
```python
setup = """
class A:
    pass
obj = A()
"""
stmt_dict ="""
data_to_set = {}
data_to_set.update({
"estAttitudeUncertQuaternionX": 0,
"estAttitudeUncertQuaternionY": 1,
"estAttitudeUncertQuaternionZ": 2,
"estAttitudeUncertQuaternionW": 3,
})
obj.__dict__.update(data_to_set)
"""
stmt_setattr = """
setattr(obj, 'estAttitudeUncertQuaternionX', 0)
setattr(obj, 'estAttitudeUncertQuaternionY', 1)
setattr(obj, 'estAttitudeUncertQuaternionZ', 2)
setattr(obj, 'estAttitudeUncertQuaternionW', 3)
"""
result = timeit.repeat(stmt=stmt_dict, setup=setup, repeat=5, number=100000)
print("Time taken by dict update: ", result)
result_2 = timeit.repeat(stmt=stmt_setattr, setup=setup, repeat=5, number=100000)
print("Time taken by setattr: ", result_2)
```
```
Time taken by dict update:  [0.031197399977827445, 0.030509116011671722, 0.030660902004456148, 0.031472728005610406, 0.02996487999917008]
Time taken by setattr:  [0.014703325025038794, 0.014561596006387845, 0.014530211017699912, 0.014202825987013057, 0.014729199989233166]
```

As you can, `setattr` is actually overall faster, even for classes without `__slots__`!

harshil21:
Ruff is failing, other changes look fine to me.

JacksonElia:
Okay I'm reverting the dict stuff, but this pr has a couple other changes I want in

harshil21 (approved):
lgtm

PR #11, opened 2024-09-17
Adjust servo extension values

harshil21:
Adjust servo `MIN/MAX` extension so it fits through the cut out


JacksonElia:
@harshil21 I don't really care if ruff fails for a testing script tbh, what do you think?

JacksonElia (approved):

PR #14, opened 2024-09-19
Clean exit on Ctrl+C

harshil21:
Ignores `SIGINT` in the child processes, so Ctrl+C events are cleanly handled within them, making the `KeyboardInterrupt` handling in `main.py` actually stop() the child processes properly.

To be merged after testing on the PI.


tests/test_airbrakes.py
```diff
@@ -1,22 +1,95 @@
...
+        assert isinstance(airbrakes.state, StandByState)
+        assert not airbrakes.shutdown_requested
+
+    def test_set_extension(self, airbrakes):
+        # Hardcoded calculated values, based on MIN_EXTENSION and MAX_EXTENSION in constants.py
+        airbrakes.set_airbrake_extension(0.5)
+        # TODO: airbrakes.current_extension must be set to 0.5 !!
+        assert airbrakes.servo.current_extension == 0.0803
+        airbrakes.set_airbrake_extension(0.0)
+        assert airbrakes.servo.current_extension == -0.0999
```
harshil21, line 39: I'm not sure if we want the servo extension to be negative, do we?

tests/test_airbrakes.py
```diff
@@ -1,22 +1,95 @@
...
+        assert airbrakes.imu == imu
+        assert airbrakes.current_extension == 0.0
+        assert isinstance(airbrakes.data_processor, IMUDataProcessor)
+        assert isinstance(airbrakes.state, StandByState)
+        assert not airbrakes.shutdown_requested
+
+    def test_set_extension(self, airbrakes):
+        # Hardcoded calculated values, based on MIN_EXTENSION and MAX_EXTENSION in constants.py
+        airbrakes.set_airbrake_extension(0.5)
+        # TODO: airbrakes.current_extension must be set to 0.5 !!
```
harshil21, line 36: Currently, `airbrakes.update()` sets `airbrakes.current_extension` to `airbrakes.servo.current_extension`, which can make the data in the logs look confusing. I propose that the value for `airbrakes.current_extension` should be detached from `servo.current_extension`.

main.py
Aman-Anas, line 1: the way I handled keyboard interrupts in the old payload code was
```py

def main(args):

    payload = kraken.PayloadSystem()

    try:
        while not payload.ready_to_shutdown:
            payload.update()

    except KeyboardInterrupt:
        pass

    payload.shutdown()

    print("Program complete. Waiting for recovery.")
```
basically just handled the keyboard interrupt in the main function. it's not like the actual payload (or in this case, airbrakes) should know anything about keyboards. Either way works though

harshil21: yeah, that also works, the main thing is to ignore SIGINT in the child processes, otherwise exceptions are still printed out and stop() isn't called for them.

I went with the context manager approach so you don't have to write boilerplate every time, e.g. in the testing scripts.

JacksonElia (changes requested):
As I was kinda saying earlier today, I think while this is the "correct" way to do this, it adds too much bloat to the code (plus looks foreign af to new programmers), so I would prefer we put a little more effort into the testing scripts rather than add this to our main code.

harshil21:
alright, reverts the context manager change for the sake of understanding code at first glance. I manually ran `scripts/run_main_local.py` and it works great.

Apparently this didn't work on the PI today, but I didn't get to much to time to see why, so hold off on merging for now.

tests/test_airbrakes.py
JacksonElia: Yes you are right. Make it so airbrakes.set_extension() sets both the servo extension and current_extension?

main.py
```diff
@@ -15,16 +15,15 @@ def main():
...
-        airbrakes.update()
-
-    # Shutdown the IMU and logger processes:
-    airbrakes.stop()
+    try:
+        airbrakes.start()  # Start the IMU and logger processes
+        # This is the main loop that will run until we press Ctrl+C
+        while not airbrakes.shutdown_requested:
+            airbrakes.update()
+    except KeyboardInterrupt:
```
JacksonElia, line 23: I don't think that a keyboard interrupt will stop the main process when the rocket is actually being. Launched, it will probably just end when we unplug the battery. That being said, we will end sims via keyboard interrupt, so I think its good.

scripts/run_main_local.py
```diff
@@ -0,0 +1,32 @@
+"""The mocked main file which can be run locally. To run this, make sure you're not inside scripts,
+and run the following command: `python -m scripts.test_imu`"""
```
JacksonElia, line 2: comment needs to be changed -- its not test_imu

JacksonElia (approved):
Looks good, just a couple small things

harshil21:
Confirmed working on the PI.

PR #15, opened 2024-09-20
Changed project structure

JacksonElia:
I wasn't happy with how our project was structured. We had too many files in airbrakes/ and I didn't like how we just had the imu folder. Also, I moved constants to the same level as main because main.py uses some constants.

To review this pr, I would just go to the branch and look at the new structure.


requirements.txt
JacksonElia, line 1: Because we have a pyprojects.toml, we don't also need a requirements.txt, right?

harshil21: technically no, but I leave that up since it's easier for beginners to grasp what the dependencies our without looking at the whole pyproject.toml file

pyproject.toml
```diff
@@ -6,6 +6,7 @@ version = "0.1.0"
 dependencies = [
     "gpiozero",
     "pigpio",  # Run sudo pigpiod before running the program
+    "pytest",
```
harshil21, line 9: no, this is already specified in "optional dependencies" below, which is for developing

harshil21 (commented):
Definitely like the new `hardware` folder. Still on the fence on whether the data packets should be in their own file, or inside `imu.py`, where they are initialized and used.

Github CI is failing because of removed requirements.txt. Should be fixed by adding `pip install .[dev]`

requirements.txt
JacksonElia: ah just saw that, thats fair. Do you think we should leave it as is, or think we should add requirements.txt back?

harshil21: I'm fine with whatever, obviously the programmer in me says to keep only one thing though

JacksonElia: alr I agree

pyproject.toml
JacksonElia: I think while its not incorrect to put it within imu.py, it would cause the file to have 4 classes and be incredibly long, so I think its good to separate it. I like it being with logger and data_processor because both of them use it

harshil21:
Can you also change the path of the logs for the `scripts`? It should be `scripts/test_logs`.

harshil21 (approved):

PR #16, opened 2024-09-20
Started writing sim code, also cleaned up IMU and State a little

JacksonElia:
What do you think of imu.py? I tried to write it well with ABC, and I think the sim should work well.


airbrakes/hardware/imu.py
```diff
@@ -123,36 +203,93 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
                 # Put the latest data into the shared queue
                 self._data_queue.put(imu_data_packet)
 
-    def get_imu_data_packet(self) -> IMUDataPacket:
-        """
-        Gets the last available data packet from the IMU.
 
-        Note: If `get_imu_data_packet` is called slower than the frequency set, the data will not
-        be the latest, but the first in the queue.
+class MockIMU(AbstractIMU):
```
harshil21, line 207: I don't see why you have to make an Abstract class just to create a `MockIMU`? Moreover this does not belong in `airbrakes/`, it should either be in `sims/` or `tests/`. 

One can just create a subclass and override any method you want, it is not necessary to create ABC's in python. 

ABC's can make sense if we have many different versions of the `IMU` class, but having a `Test/MockIMU` should not necessiate the change.

airbrakes/hardware/imu.py
```diff
@@ -123,36 +203,93 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
+                    imu_data_packet.estOrientQuaternionX = float(row.get("estOrientQuaternionX"))
+                    imu_data_packet.estOrientQuaternionY = float(row.get("estOrientQuaternionY"))
+                    imu_data_packet.estOrientQuaternionZ = float(row.get("estOrientQuaternionZ"))
+                    imu_data_packet.estPressureAlt = float(row.get("estPressureAlt"))
+                    imu_data_packet.estAttitudeUncertQuaternionW = float(row.get("estAttitudeUncertQuaternionW"))
+                    imu_data_packet.estAttitudeUncertQuaternionX = float(row.get("estAttitudeUncertQuaternionX"))
+                    imu_data_packet.estAttitudeUncertQuaternionY = float(row.get("estAttitudeUncertQuaternionY"))
+                    imu_data_packet.estAttitudeUncertQuaternionZ = float(row.get("estAttitudeUncertQuaternionZ"))
+
+                if imu_data_packet is None:
```
harshil21, line 288: possible unbound variable

airbrakes/state.py
```diff
@@ -59,11 +59,9 @@ class StandByState(State):
 
     __slots__ = ()
 
-    @override
```
harshil21, line 62: why?

airbrakes/hardware/imu.py
```diff
@@ -123,36 +203,93 @@ def _fetch_data_loop(self, port: str, frequency: int, _: bool):
...
+                    imu_data_packet.estAngularRateZ = float(row.get("estAngularRateZ"))
+                    imu_data_packet.estOrientQuaternionW = float(row.get("estOrientQuaternionW"))
+                    imu_data_packet.estOrientQuaternionX = float(row.get("estOrientQuaternionX"))
+                    imu_data_packet.estOrientQuaternionY = float(row.get("estOrientQuaternionY"))
+                    imu_data_packet.estOrientQuaternionZ = float(row.get("estOrientQuaternionZ"))
+                    imu_data_packet.estPressureAlt = float(row.get("estPressureAlt"))
+                    imu_data_packet.estAttitudeUncertQuaternionW = float(row.get("estAttitudeUncertQuaternionW"))
+                    imu_data_packet.estAttitudeUncertQuaternionX = float(row.get("estAttitudeUncertQuaternionX"))
+                    imu_data_packet.estAttitudeUncertQuaternionY = float(row.get("estAttitudeUncertQuaternionY"))
+                    imu_data_packet.estAttitudeUncertQuaternionZ = float(row.get("estAttitudeUncertQuaternionZ"))
```
harshil21, line 286: why not a for loop

harshil21 (changes requested):
Sim code should be in another directory, also merge main please

JacksonElia:
Wait wdym merge main?

airbrakes/hardware/imu.py
JacksonElia: cause this was easy and each row has more data than just whats in the packet

airbrakes/state.py
JacksonElia: When looking on the docs and various stack overflow posts, I never saw people using @override with ABC. It's just not necessary/part of ABC

harshil21:
> Wait wdym merge main?

the commits from main branch are being shown in the diff. If you merge main branch into yours, it will be fixed, after conflict resolution ofc

airbrakes/hardware/imu.py
JacksonElia: I want both IMU and MockIMU to have the same superclass, so how would you recommend having the file structure if we did it that way? Putting it all in the same file was the cleanest way I could think but idk

harshil21: keep the IMU class as is. `MockIMU` can go in `sims/`, and just inherit from `IMU`. You can override any of the methods you want, e.g.

```python
class MockIMU(IMU):
    def _fetch_data_loop(self, *args, **kwargs):
        # logic for reading file and putting packets in queue here
```

From what I can see, the only change you want in the MockIMU class is to read data from the csv file as packets. This is similar to what I did in the tests, see e.g. `test_data_packets_fetch` in `test_imu.py`. In that test, I didn't even need to subclass. Depending on how much you want to change, subclassing may be a better option. However, the main files inside `airbrakes` must remain unchanged (unless you see a bug or something ofc)

JacksonElia: Okay that sounds good

airbrakes/hardware/imu.py
```diff
@@ -34,7 +36,12 @@ class IMU:
         "_running",
     )
 
-    def __init__(self, port: str, frequency: int, upside_down: bool):
+    def __init__(self, port: str, frequency: int):
+        """
+        Initializes the object that interacts with the physical IMU connected to the pi.
+        :param port: the port that the IMU is connected to
+        :param frequency: the frequency in Hz that the IMU will be polled at
```
harshil21, line 43: technically no, we don't control the frequency at all from this code

airbrakes/hardware/imu.py
```diff
@@ -43,26 +50,61 @@ def __init__(self, port: str, frequency: int, upside_down: bool):
...
+            data_packet = self.get_imu_data_packet()
+            data_packets.append(data_packet)
+
+        return data_packets
+
+    def _fetch_data_loop(self, port: str, frequency: int) -> None:
+        """
+        The loop that fetches data from the IMU. It runs in parallel with the main loop.
+        :param port: the port that the IMU is connected to
+        :param frequency: the frequency in Hz that the IMU will be polled at
```
harshil21, line 107: .

airbrakes/mock/mock_imu.py
harshil21, line 1: i'd still prefer `mock/` to be outside `airbrakes/`, i.e. in the root project directory

JacksonElia: I feel like it should be within airbrakes/ because it is essentially acting as a subsystem on the rocket.

JacksonElia:
I want to test this before merging but it should work

harshil21 (approved):

PR #19, opened 2024-09-22
State determination

harshil21:
Implements state determination and transition. 

Additional changes:

- IMUDataProcessor class now calculates speed, max_speed, and zeros out the altitude at init
- Deploys airbrakes after Motor burnout.
- Added `LandedState`
- Removes `run_main_local` file since we can now run the main file in mock mode.
- Adds 2 csv files


airbrakes/data_handling/data_processor.py
```diff
@@ -87,6 +79,138 @@ def avg_acceleration_mag(self) -> float:
...
+        return avg_x_accel, avg_y_accel, avg_z_accel
+
+    def _calculate_speed(self, a_x: list[float], a_y: list[float], a_z: list[float]) -> float:
+        """
+        Calculates the speed of the rocket based on the linear acceleration.
+        Integrates the linear acceleration to get the speed.
+        """
+        # We need at least two data points to calculate the speed:
+        if len(self._data_points) < 2:
+            return 0.0
```
harshil21, line 194: I'd still like us to reconsider fetching a set amount of data packets in `get_imu_data_packets()`, so that this condition is never hit and things are a little more consistent overall.

harshil21 (commented):
Confidence is slowly increasing that our system will indeed work. A full scale integration test (simulation) should iron out the last remaining bugs / quirks.

JacksonElia (changes requested):

airbrakes/data_handling/data_processor.py
```diff
@@ -89,6 +81,138 @@ def avg_acceleration_mag(self) -> float:
...
```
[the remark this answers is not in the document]
harshil21: done in #22

airbrakes/data_handling/data_processor.py
harshil21: we can tell it, "give me at least 10 packets of data, but if you have more, give it all"

JacksonElia (approved):

JacksonElia (approved):

PR #24, opened 2024-09-26
Fix speed calculations - approach 1

harshil21:
Need to test:

- How the speed graph looks like after apogee for winter '23 flight data
- Running sim for other flight datas and comparing speed


JacksonElia (approved):

harshil21:
tests failing tho

PR #26, opened 2024-09-27
Reduce log file sizes, add `ProcessedDataPacket`, rename `FreeFallState`

JacksonElia:
- Rename FreeFallState to CoastState
- Reduce file log size by truncating decimal places to 8, and removing more unecessary fields
- Add `ProcessedDataPacket` and `LoggedDataPacket` to make it more clear what is being logged and how.
- Airbrakes now deploy 1.5 seconds after entering `CoastState`.

PR #28, opened 2024-09-27
Add log buffer 

harshil21:
- Logs 15000 packets max when in StandbyState and LandedState
- Buffer is also implemented when the state changes from StandbyState to MotorBurnState, so we don't lose any logs (buffer is of a 1000 packets, and the pi on average processes about 25 packets.)

Can be merged to #26


harshil21:
Logger tests were created and adapted to new code and they pass.

harshil21:
Log buffer works perfectly with mock sim data.

PR #31, opened 2024-09-30
fixed element order of quaternion vectors

wlsanderson:
Changed quaternions so they are logged as [w x y z], we had it as [x y z w], which is out of order.


JacksonElia (approved):
LGTM

harshil21 (approved):

PR #32, opened 2024-10-03
Log and use invalid data packets

harshil21:
As discussed by @wlsanderson and I, the interest launch did not log all quaternions since they were marked "invalid" by the IMU - Why?

Due to high rotation rates during coast and free fall, the IMU cannot reliably process data even though it was well within the specifications of the IMU, and there are no threshold settings we can change with regards to this. 

Thus, we decided to log invalid data packets as well. This can be used in the apogee prediction, which would have used the noisy/inaccurate data anyway, but using the quaternions directly saves us some computation.

PR changes:

- All attributes of a data packet are set, whether it is valid or not.
- Adds the `invalid_fields` parameter to identify which fields of the data packet were invalid

### DIFF
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -92,9 +92,7 @@ class AirbrakesContext:
         i = 0
         for data_packet in data_packets:
             logged_data_packet = LoggedDataPacket(
-                state=self.state.name[0],
-                extension=self.current_extension.value,
-                timestamp=data_packet.timestamp
+                state=self.state.name[0], extension=self.current_extension.value, timestamp=data_packet.timestamp
             )
             logged_data_packet.set_imu_data_packet_attributes(data_packet)
             if isinstance(data_packet, EstimatedDataPacket):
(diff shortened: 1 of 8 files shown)


airbrakes/airbrakes.py
```diff
@@ -92,7 +92,8 @@ def update(self) -> None:
         i = 0
         for data_packet in data_packets:
             logged_data_packet = LoggedDataPacket(
-                state=self.state.name[0], extension=self.current_extension.value, timestamp=data_packet.timestamp
```
JacksonElia, line 95: where are you setting timestamp? Because timestamp has no initial value, I thought it has to be set upon creation

harshil21: `set_imu_data_packet_attributes` automatically does this, there is no need of duplication here. I will add a specific test for this in another PR.

harshil21: ugh, you might be right. So much for trying to separate my PRs.

JacksonElia (approved):

PR #33, opened 2024-10-03
Add more fields to data packets (needs to be added w/ sensor connect)

JacksonElia:


airbrakes/data_handling/imu_data_packet.py
```diff
@@ -26,6 +26,18 @@ class RawDataPacket(IMUDataPacket):
...
+    deltaVelY: float | None = None
+    deltaVelZ: float | None = None
+    # TODO: check if these exist
+    deltaThetaX: float | None = None
+    deltaThetaY: float | None = None
+    deltaThetaZ: float | None = None
+    # mag units are in gauss, they can be used in addition to the accelerometer to determine orientation
+    stabilizedMagX: float | None = None
+    stabilizedMagY: float | None = None
+    stabilizedMagZ: float | None = None
```
harshil21, line 40: pretty sure these do not exist. We do not have a magnetometer.

JacksonElia: they will need to be tested, but sensor connect said it did

harshil21 (changes requested):
The new fields need to be added to `LoggedDataPacket`.

harshil21 (approved):

PR #35, opened 2024-10-03
Fix and add many tests

harshil21:
PR changes:
- Adds and fixes all tests
- Fixes all mock sims
- Adds integration test, which runs on interest launch data and winter_2023_launch data.
- Check code coverage: 86%. imu and mock_imu.py are mostly the ones not being covered.

Not doing: 

- Add bash script to auto install mscl depending on platform
- Add mscl installation to CI
- Add windows runner to the CI?
- Check if we can mock the mscl library in test_imu to fully test `_fetch_data_loop`

### DIFF
--- a/airbrakes/utils.py
+++ b/airbrakes/utils.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
 MOVE_CURSOR_UP = "\033[F"  # Move the cursor one line up
 
 
-def convert_to_nanoseconds(value) -> int:
+def convert_to_nanoseconds(value: str | int) -> int:
     """Converts seconds to nanoseconds, if `value` is in float."""
     try:
         return int(float(value) * 1e9)
@@ -18,7 +18,7 @@ def convert_to_nanoseconds(value) -> int:
         return None
 
 
-def convert_to_float(value) -> float | None:
+def convert_to_float(value: str | int) -> float | None:
     """Converts a value to a float, returning None if the conversion fails."""
     try:
         return float(value)  # Attempt to convert to float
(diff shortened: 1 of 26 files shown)


JacksonElia:
TODO looks good except for: I don't think we need to auto install mscl--it's a lot of work, plus the imu won't even work for our laptops locally, plus I don't think the imu class will be changed that much in the future.

harshil21:
> TODO looks good except for: I don't think we need to auto install mscl--it's a lot of work, plus the imu won't even work for our laptops locally, plus I don't think the imu class will be changed that much in the future.

Okay. It's just that from a python packaging perspective, it makes sense to install the dependencies from one place. Tests for this will probably be...long.

airbrakes/utils.py
```diff
@@ -10,23 +10,28 @@
 MOVE_CURSOR_UP = "\033[F"  # Move the cursor one line up
 
 
-def convert_to_nanoseconds(value) -> int:
-    """Converts seconds to nanoseconds, if `value` is in float."""
+def convert_to_nanoseconds(timestamp_str: str) -> int | None:
```
JacksonElia, line 13: maybe we could do a more sophisticated check? e.g. check if the number passed into here is more than 10^9?

harshil21: yeah perhaps. But the better solution is to just clean the csv data in the first place

JacksonElia:
LGTM so far. Can we try to merge this soon? Also, can we merge #37 into this?

harshil21:
@JacksonElia Yeah, I'm just waiting for you to write tests in #37 so we can merge that into this and then merge this.

JacksonElia (approved):
Lgtm

PR #40, opened 2024-10-09
Performance Improvements

harshil21:
Improves performance of the main process by about 25% by simply inlining calls, removing unused calculations (average_acceleration), and minimizing usage of `isinstance`, `getattr()`, `setattr()`.

Here is the profiled svg flowchart generated by [`pytest-profiling`](https://pypi.org/project/pytest-profiling/) when running `test_integration.py`. (You should `pip install graphviz` as well if you want to run that locally)

Open the image in a new tab to be able to zoom and scroll:

### Before this change:
combined

Time taken to run `pytest test_integration.py` once: 16.98s

### After this change:
combined

Time taken to run `pytest test_integration.py` once: 12.71s

Overall improvement from running `pytest test_integration.py`: **25.14%**

### DIFF
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -91,15 +91,15 @@ class AirbrakesContext:
         # data packet
         i = 0
         for data_packet in data_packets:
-            logged_data_packet = LoggedDataPacket(
+            ldp = LoggedDataPacket(
                 state=self.state.name[0], extension=self.current_extension.value, timestamp=data_packet.timestamp
             )
-            logged_data_packet.set_imu_data_packet_attributes(data_packet)
+            ldp.set_imu_data_packet_attributes(data_packet)
             if isinstance(data_packet, EstimatedDataPacket):
-                logged_data_packet.set_processed_data_packet_attributes(processed_data_packets[i])
+                ldp.set_processed_data_packet_attributes(processed_data_packets[i])
                 i += 1
 
-            logged_data_packets.append(logged_data_packet)
+            logged_data_packets.append(ldp)
         # Logs the current state, extension, IMU data, and processed data
         self.logger.log(logged_data_packets)
 
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -91,15 +91,15 @@ class AirbrakesContext:
         # data packet
         i = 0
         for data_packet in data_packets:
-            ldp = LoggedDataPacket(
+            logged_data_packet = LoggedDataPacket(
                 state=self.state.name[0], extension=self.current_extension.value, timestamp=data_packet.timestamp
             )
-            ldp.set_imu_data_packet_attributes(data_packet)
+            logged_data_packet.set_imu_data_packet_attributes(data_packet)
             if isinstance(data_packet, EstimatedDataPacket):
-                ldp.set_processed_data_packet_attributes(processed_data_packets[i])
+                logged_data_packet.set_processed_data_packet_attributes(processed_data_packets[i])
                 i += 1
 
-            logged_data_packets.append(ldp)
+            logged_data_packets.append(logged_data_packet)
         # Logs the current state, extension, IMU data, and processed data
         self.logger.log(logged_data_packets)
 
--- a/airbrakes/data_handling/logged_data_packet.py
+++ b/airbrakes/data_handling/logged_data_packet.py
@@ -2,9 +2,8 @@
 
 import msgspec
 
-from airbrakes.data_handling.imu_data_packet import IMUDataPacket
+from airbrakes.data_handling.imu_data_packet import EstimatedDataPacket, IMUDataPacket, RawDataPacket
 from airbrakes.data_handling.processed_data_packet import ProcessedDataPacket
-from constants import DATA_PACKET_DECIMAL_PLACES
 
 
 class LoggedDataPacket(msgspec.Struct):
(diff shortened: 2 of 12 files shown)


airbrakes/airbrakes.py
```diff
@@ -91,15 +91,15 @@ def update(self) -> None:
         # data packet
         i = 0
         for data_packet in data_packets:
-            logged_data_packet = LoggedDataPacket(
+            ldp = LoggedDataPacket(
```
JacksonElia, line 94: I don't think acronyms are every good var names, it makes it less clear

JacksonElia: and this isn't just me, I read this dude's book and it's really good. I highly recommend. https://www.google.com/books/edition/Clean_Code/_i6bDeoCQzsC?hl=en&gbpv=1&printsec=frontcover

airbrakes/data_handling/logged_data_packet.py
```diff
@@ -59,30 +58,101 @@ class LoggedDataPacket(msgspec.Struct):
...
-        for key in imu_data_packet.__struct_fields__:
-            if hasattr(self, key):
-                value = getattr(imu_data_packet, key)
-                # Only logs the 8 decimal places as there is already so much noise in the data
-                if isinstance(value, float):
-                    value = round(value, DATA_PACKET_DECIMAL_PLACES)
-                setattr(self, key, value)
-            else:
-                raise AttributeError(f"{key} is not a valid attribute")
+        idp = imu_data_packet
```
JacksonElia, line 75: same here

airbrakes/data_handling/logged_data_packet.py
```diff
@@ -59,30 +58,101 @@ class LoggedDataPacket(msgspec.Struct):
...
+        if isinstance(idp, EstimatedDataPacket):
+            # super ugly code, but it results in a 10-14% speedup overall
```
JacksonElia, line 77: lol python moment that this is what we have to do -- side note, I thought you said that with structs, setattr was actually just as fast? Does that mean we need to change the stuff for raw and est data packets too?

airbrakes/hardware/imu.py
```diff
@@ -96,9 +96,8 @@ def get_imu_data_packets(self) -> collections.deque[IMUDataPacket]:
         # We use a deque because it's faster than a list for popping from the left
         data_packets = collections.deque()
         # While there is data in the queue, get the data packet and add it to the dequeue which we return
-        while not self._data_queue.empty():
-            data_packet = self.get_imu_data_packet()
-            data_packets.append(data_packet)
+        for _ in range(self._data_queue.qsize()):
```
JacksonElia, line 99: why change the for loop to a while loop?

JacksonElia (approved):
Just some minor nits and questions, otherwise, good stuff!

airbrakes/data_handling/logged_data_packet.py
harshil21: > lol python moment that this is what we have to do

😭 hopefully future versions of python resolves that

> -- side note, I thought you said that with structs, setattr was actually just as fast?

setattr is faster than adding and updating a dictionary with/without slots and structs, but apparently nothing beats directly setting attributes on an object.

> Does that mean we need to change the stuff for raw and est data packets too?

that isn't a bottleneck since it's on its own process. You can verify that by printing queue size in the main process and you'll mostly see a number > 0 (w/ the real IMU). Though it has been a while since I tested that, we should try it again today.

airbrakes/hardware/imu.py
harshil21: technically it saves repeated calls to .empty() every iteration

airbrakes/data_handling/logged_data_packet.py
```diff
@@ -59,30 +58,101 @@ class LoggedDataPacket(msgspec.Struct):
...
+            if idp.estGravityVectorX is not None:
+                self.estGravityVectorX = f"{idp.estGravityVectorX:.8f}"
+            if idp.estGravityVectorY is not None:
+                self.estGravityVectorY = f"{idp.estGravityVectorY:.8f}"
+            if idp.estGravityVectorZ is not None:
+                self.estGravityVectorZ = f"{idp.estGravityVectorZ:.8f}"
+            if idp.estPressureAlt is not None:
+                self.estPressureAlt = f"{idp.estPressureAlt:.8f}"
+
+        if isinstance(idp, RawDataPacket):
```
harshil21, line 124: could be an else statement and save another call to isinstance

airbrakes/data_handling/logged_data_packet.py
```diff
@@ -59,30 +58,101 @@ class LoggedDataPacket(msgspec.Struct):
...
-            else:
-                raise AttributeError(f"{key} is not a valid attribute")
+        idp = imu_data_packet
+        if isinstance(idp, EstimatedDataPacket):
+            # super ugly code, but it results in a 10-14% speedup overall
+            # The speed improvements come from not looping through the fields of the data packet
+            # and using getattr() and setattr() to set the attributes of the logged data packet.
+            # Additionally, rounding using f-strings is faster than round() by about ~25%
+            if idp.estOrientQuaternionW is not None:
+                self.estOrientQuaternionW = f"{idp.estOrientQuaternionW:.8f}"
```
harshil21, line 82: doing something like `(attr:=idp.estOrientQuaternionW)` and just using `attr:.8f` could provide another slight speedup since we save another unnecessary attribute access. Need to profile that and check

airbrakes/data_handling/logger.py
```diff
@@ -74,20 +76,22 @@ def stop(self) -> None:
...
         Logs the current state, extension, and IMU data to the CSV file.
+
         :param logged_data_packets: the list of IMU data packets to log
         """
         # Loop through all the IMU data packets
         for logged_data_packet in logged_data_packets:
             # Formats the log message as a CSV line
-            message_dict = {key: getattr(logged_data_packet, key) for key in logged_data_packet.__struct_fields__}
+            # We are populating a dictionary with the fields of the logged data packet
+            message_dict = msgspec.structs.asdict(logged_data_packet)
```
harshil21, line 86: we should probably just import asdict, and then use that instead of accessing the attribute every iteration

airbrakes/hardware/imu.py
harshil21, line 99: a list comprehension is likely going to be faster now. Will profile and check.

harshil21 (commented):
looks like we can squeeze the lemon some more

airbrakes/data_handling/logged_data_packet.py
harshil21: looks like doing attr:= actually takes slightly more time, about 0.25% slower

airbrakes/hardware/imu.py
harshil21: not any better.

airbrakes/data_handling/logged_data_packet.py
harshil21: @JacksonElia ?

JacksonElia: I still think it shouldn't be abbreviated

PR #41, opened 2024-10-12
Added method for calculating speed from altitude differences

JacksonElia:


airbrakes/hardware/imu.py
```diff
@@ -160,6 +160,10 @@ def _fetch_data_loop(self, port: str, frequency: int) -> None:
                                 setattr(imu_data_packet, f"{channel}X", matrix.as_floatAt(0, 1))
                                 setattr(imu_data_packet, f"{channel}Y", matrix.as_floatAt(0, 2))
                                 setattr(imu_data_packet, f"{channel}Z", matrix.as_floatAt(0, 3))
+                            case "estPressureAlt":
+                                # Converts the pressure altitude to meters and sets it to the data packet object
+                                # TODO: We can maybe change this in sensor connect
+                                setattr(imu_data_packet, channel, data_point.as_float() * FEET_TO_METERS)
```
harshil21, line 166: It is in meters already? E.g. see your graph of the interest launch in the readme, and then Donald's attached picture of the flight data from the altimeter in slack

airbrakes/data_handling/data_processor.py
```diff
@@ -160,15 +164,32 @@ def get_processed_data(self) -> list[ProcessedDataPacket]:
...
+        speeds_from_altitude_interpolated = []
+        for i in range(len(self._current_altitudes)):
+            # If the altitude has changed, we need to interpolate the speed, but you get n - 1 speeds from n altitudes,
+            # so we need to check that we have speeds left
+
+            if self._current_altitudes[i] != self._previous_altitude and len(self._speeds_from_altitude) != 0:
+                self._previous_altitude = self._current_altitudes[i]
+                self._previous_velocity_from_altitude = self._speeds_from_altitude[0]
+                self._speeds_from_altitude = self._speeds_from_altitude[1:]
+            speeds_from_altitude_interpolated.append(self._previous_velocity_from_altitude)
```
harshil21, line 179: I suspected that there was a numpy function for interpolation so I asked chatgpt and it looks like there is, not sure if you've looked at that already:

```python
import numpy as np

# Example altitudes and speeds (make sure to replace with your actual data)
altitudes = np.array(self._current_altitudes)
speeds = np.array(self._speeds_from_altitude)

# Interpolating the speeds for the entire altitudes range
speeds_interpolated = np.interp(altitudes, self._previous_altitudes, self._previous_speeds_from_altitude)

# If you need to append the interpolated speeds to a list or similar structure
speeds_from_altitude_interpolated = list(speeds_interpolated)
```

airbrakes/data_handling/data_processor.py
```diff
@@ -245,7 +268,33 @@ def _calculate_speeds(self, a_x: list[float], a_y: list[float], a_z: list[float]
...
+        # Accounts for start up not having a last data point
+        if self._last_data_point is None:
+            return np.array([0.0] * len(self._data_points))
+
+        # Because the pressure altitude is only updated at 50Hz, we have to check if the data point is new, we do this
+        # by assuming that we will almost never get the same altitude reading twice in a row, so if we get the unique
+        # altitudes, that should be each altitude reading. In theory, we don't actually need to do this as the math will
+        # work out, but it's better to be safe than sorry and I like how this is more explict/intentional.
+        altitude_list.append(self._last_data_point.estPressureAlt)
+        unique_altitudes = np.unique(altitude_list)
```
harshil21, line 290: do not modify the original passed in list (pressure_altitudes). Always make a new copy. Also, why is the last data point of the pressure alt added to the end instead of the beginning?

airbrakes/data_handling/data_processor.py
```diff
@@ -245,7 +268,33 @@ def _calculate_speeds(self, a_x: list[float], a_y: list[float], a_z: list[float]
...
+
+        # We need at least two data points to calculate the speed:
+        if len(unique_altitudes) < 2:
+            return np.array([0.0])
+
+        # TODO: calculate dt in a more sophisticated way
+        time_diff = 1 / 50.0
```
harshil21, line 297: we need to process the data packets from the source directly to get the corresponding dt along with pressure alt.

airbrakes/data_handling/processed_data_packet.py
```diff
@@ -9,6 +9,8 @@ class ProcessedDataPacket(msgspec.Struct):
     data.
     """
 
-    avg_acceleration: tuple[float, float, float]
-    current_altitude: float
-    speed: float
+    # pro for processed, I don't like the abbreviation, but it's better consistent est
```
harshil21, line 12: "consistent est" consistent what now? Also when I read "pro" initially I had no idea what that was supposed to mean lol

harshil21 (changes requested):
draft review

airbrakes/data_handling/data_processor.py
JacksonElia: that linearly interpolates rather than doing it in steps. Both are fine as we will just be using the actual values in the kalman filter, not the interpolated ones. I feel like stepwise interpolation is better tho as it is more accurate to what is actually happening

airbrakes/hardware/imu.py
JacksonElia: i'm tweaking I guess

JacksonElia: no my log file that I was using (idk which one lol) is in feet. I'll keep it as an option and set if false by default

PR #43, opened 2024-10-14
Add cpu usage in mock simulation

harshil21:
Allows us to inspect CPU usage to see if any process is being held up while running the mock sim (in real time).

On my PC it shows 0% for all of them most of the time, with some spikes happening mostly for the MockIMU (just reading the csv file)

Will be useful to run on the pi and check.

image


JacksonElia (approved):
Awesome! I love your display!

PR #47, opened 2024-10-17
Implement Apogee Prediction

harshil21:
Major changes:

- Implementation of vertical velocity using rotated acceleration in the earth's frame of reference. This is more accurate than our current magnitude approach. However it is still slightly inaccurate during free fall by about 2-5m/s (just eyeballing it v/s altitude).

- Apogee Prediction using curve fitting. This runs only for the duration of CoastPhase. We converge to true apogee with an accuracy of +/- 10m within 5 seconds of coasting. 

- The apogee prediction process runs only every 10 packets.

WIP:

- [x] Tests (verify that everything works as expected)
- [x] multiprocessing
- [ ] Review comments
- ~plot velocity calculated from rotated frame vs speed from accel~


airbrakes/data_handling/apogee_prediction.py
```diff
@@ -62,11 +66,12 @@ def __init__(self, state: State, data_processor: IMUDataProcessor, context: "Air
         self._time_diff: npt.NDArray[np.float64] | None = None
 
         self._gravity = 9.798  # will use gravity vector in future
-        # self._data_points = multiprocessing.Queue(maxsize=100)
-        # self._running = multiprocessing.Value("b", False)  # Makes a boolean value that is shared between processes
-        # self._update_process = multiprocessing.Process(
-        #     target=self.update, name="Apogee Prediction Process"
-        # )
+        # TODO: decide on packet limit? Also how does it work? If we try to add 200 packets and the limit is 100,
+        #  will it add the first 100 packets or the last 100 packets (we would want last 100--this could be a time save)
```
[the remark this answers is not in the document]
JacksonElia: okay let's come back to this later once things are working/we need performance. For now, I'm going to do no limit.

airbrakes/airbrakes.py
```diff
@@ -101,14 +101,15 @@ def update(self) -> None:
...
         self.data_processor.update(est_data_packets)
-        if self.state.name[0] == "C":  # Only run apogee prediction in coast state:
-            pass
-            # self.apogee_predictor.update(est_data_packets)
 
         # Get the processed data packets from the data processor, this will have the same length as the number of
         # EstimatedDataPackets in data_packets
         processed_data_packets: deque[ProcessedDataPacket] = self.data_processor.get_processed_data_packets()
 
+        if isinstance(self.state, CoastState):  # Only run apogee prediction in coast state:
```
harshil21, line 109: I deliberately didn't use `isinstance` for performance reasons. Not pythonic, but just pre-optimizing based on previous profiling. I'll specifically profile line by line later.

JacksonElia: okay I'll revert.

JacksonElia: does it actually matter thaaat much though? If so, we should see If we can optimize this further
```py
        est_data_packets = [
            data_packet
            for data_packet in imu_data_packets.copy()
            if isinstance(data_packet, EstimatedDataPacket)
            # The copy() above is critical to ensure the data here is not modified by the data processor
        ]
```

harshil21:
last working commit before multiprocessing - https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/commit/0ac977ce9afd6469e94fc4832ad11a6759899414 (I think).

harshil21:
With the new and accurate CPU usage in the Flight Display, we can actually see how much CPU usage Apogee Prediction takes. And it's surprisingly the least amount of compute out of all the processes, averaging about 0.3% on PC during `CoastState`.

harshil21:
Velocity for purple nurple is very wrong during Free fall because we don't have accurate 3D data for the launch. That launch was missing Angular rate and quaternions, and scaledGyro isn't accurate enough for a 3D representation. 

So once we are in free fall, the rocket is rotating a lot, and we can't accurately quantify it, because our rotation data is only accurate for 2D.

airbrakes/data_handling/data_processor.py
```diff
@@ -132,74 +202,100 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+            # For simplicity, we multiply by -1 so that acceleration during motor burn is positive, and
+            # acceleration due to drag force during coast phase is negative.
+            rotated_accelerations[idx] = accel_rotated_quat[3] * -1
+
+        return rotated_accelerations
+
+    def _calculate_vertical_velocity(self) -> npt.NDArray[np.float64]:
+        """
+        Calculates the velocity of the rocket based on the linear acceleration. Integrates the
+        linear acceleration to get the velocity.
```
harshil21, line 269:
```suggestion
        Calculates the velocity of the rocket based on the rotated compensated acceleration. Integrates
        that acceleration to get the velocity.
```

airbrakes/data_handling/data_processor.py
```diff
@@ -132,74 +202,100 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-        # All the speeds gotten as the magnitude of the velocity vector at each point
-        return np.sqrt(velocities_x**2 + velocities_y**2 + velocities_z**2)
+        return vertical_velocities
 
-    def _get_time_differences(self) -> npt.NDArray[np.float64]:
+    def _calculate_time_differences(self) -> npt.NDArray[np.float64]:
         """
         Calculates the time difference between each data point and the previous data point. This cannot
-        be called on the first update as _last_data_point is None.
+        be called on the first update as _last_data_packet is None.
```
harshil21, line 294:
```suggestion
```

airbrakes/data_handling/data_processor.py
```diff
@@ -132,74 +202,100 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+        self._previous_vertical_velocity = vertical_velocities[-1]
 
-        # All the speeds gotten as the magnitude of the velocity vector at each point
-        return np.sqrt(velocities_x**2 + velocities_y**2 + velocities_z**2)
+        return vertical_velocities
 
-    def _get_time_differences(self) -> npt.NDArray[np.float64]:
+    def _calculate_time_differences(self) -> npt.NDArray[np.float64]:
         """
         Calculates the time difference between each data point and the previous data point. This cannot
```
harshil21, line 293:
```suggestion
        Calculates the time difference between each data packet and the previous data packet.
```

airbrakes/data_handling/data_processor.py
```diff
@@ -132,74 +202,100 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-        be called on the first update as _last_data_point is None.
+        be called on the first update as _last_data_packet is None.
         :return: A numpy array of the time difference between each data point and the previous data point.
         """
         # calculate the time differences between each data point
```
harshil21, line 297:
```suggestion
        # calculate the time differences between each data packet
```

airbrakes/data_handling/data_processor.py
```diff
@@ -132,74 +202,100 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-        # We are converting from ns to s, since we don't want to have a speed in m/ns^2
-        # We are using the last data point to calculate the time difference between the last data point from the
+        # We are converting from ns to s, since we don't want to have a velocity in m/ns^2
+        # We are using the last data point to calculate the time difference  between the last data point from the
```
harshil21, line 299:
```suggestion
        # We are using the last data point to calculate the time difference between the last data point from the
```

airbrakes/state.py
```diff
@@ -113,18 +112,21 @@ def update(self):
...
         # Fallback: if our motor has burned for longer than its burn time, go to the next state
-        if time.time() - self.start_time > MOTOR_BURN_TIME:
-            self.next_state()
-            return
+
+        # =======DISABLED==========
+        # if time.time() - self.start_time > MOTOR_BURN_TIME:
+        #     print('fallback switch')
+        #     self.next_state()
+        #     return
```
harshil21, line 129: Do we want to enable this back? I think we should, since we do need a fallback condition

constants.py
```diff
@@ -46,11 +46,11 @@ class ServoExtension(Enum):
 SIMULATION_MAX_QUEUE_SIZE = 15
 
 # -------------------------------------------------------
-# Orientation Configuration
+# Data Processing Configuration
 # -------------------------------------------------------
 
 # Should be checked before launch
-UPSIDE_DOWN = False
+Z_DOWN = (1, 1, -1)
```
harshil21, line 53:
```suggestion
```

constants.py
```diff
@@ -46,11 +46,11 @@ class ServoExtension(Enum):
 SIMULATION_MAX_QUEUE_SIZE = 15
 
 # -------------------------------------------------------
-# Orientation Configuration
+# Data Processing Configuration
 # -------------------------------------------------------
 
 # Should be checked before launch
```
harshil21, line 52:
```suggestion
```

tests/imu_data/xminus.csv
harshil21, line 1: We need a readme in this folder explaining what these are

airbrakes/state.py
```diff
@@ -146,13 +148,20 @@ def __init__(self, context: "AirbrakesContext"):
     def update(self):
         """Checks to see if the rocket has reached apogee, indicating the start of free fall."""
 
-        # We extend the airbrakes after 1.5 seconds of coasting:
-        if time.time() - self.start_time > AIRBRAKES_AFTER_COASTING and not self.airbrakes_extended:
+        # Check if we are going to overshoot our target apogee, and extend the airbrakes if we are.
+        pred_apogee = self.context.apogee_predictor.apogee
+
+        if pred_apogee is not None and pred_apogee >= TARGET_ALTITUDE:
```
harshil21, line 154: Is there any concern of our predicted apogee being wildly high during first few hundred data packets?

wlsanderson: Yeah its definitely possible, but it's more likely the other way around. The first batch of data points could have some weird noise in it that causes the curve fit to estimate lower drag, which would cause the prediction to shoot high.

airbrakes/state.py
wlsanderson: We should, but we need to figure out what our motor burn time is, and add maybe a quarter second to compensate for the velocity to reach 0.97*maxVelocity

wlsanderson: It will be much safer to have the motor_burn_time constant be a higher number rather than lower. Switching too early could lead to apogee prediction converging 3 or 4 seconds later, or worse, not converging at all. If we switch too late, we might lose a half-second or second of coast phase, but it will still converge quickly.

harshil21: ok. But if we are going with that, I'd like us to use the timestamp from the data packets, and not system time, since we don't want to include cpu run time.

airbrakes/state.py
harshil21: So it might be wise to still use a time based check on this as well, e.g. check for this condition only after 2 seconds, so it's converged

airbrakes/airbrakes.py
```diff
@@ -43,15 +52,17 @@ def __init__(self, servo: Servo, imu: IMU, logger: Logger, data_processor: IMUDa
...
         self.servo = servo
         self.imu = imu
         self.logger = logger
         self.data_processor = data_processor
+        self.apogee_predictor = apogee_predictor
 
         # Placeholder for the current airbrake extension until they are set
         self.current_extension: ServoExtension = ServoExtension.MIN_EXTENSION
-
+        # The rocket starts in the StandByState
```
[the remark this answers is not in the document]
JacksonElia: good eye, I'll refactor

airbrakes/airbrakes.py
```diff
@@ -150,3 +147,10 @@ def retract_airbrakes(self) -> None:
         """
         self.servo.set_retracted()
         self.current_extension = ServoExtension.MIN_EXTENSION
+
+    def predict_apogee(self) -> None:
+        """
+        Predicts the apogee of the rocket based on the current processed data. This
+        should only be called in the coast state, before we start controlling the airbrakes.
+        """
+        self.apogee_predictor.update(self.processed_data_packets)
```
harshil21, line 156: oh yeah this is definitely cleaner, nice job

wlsanderson:
we so close

tests/imu_data/xminus.csv
JacksonElia: @wlsanderson

airbrakes/state.py
JacksonElia: I'm deleting this code for now and we will come back to control in another PR after this is merged. This code won't work because the curve fit is only for data without any airbrakes extension, and as soon as we extend them it will go to shit

harshil21: :(

That will fail the integration test and a bunch of unit tests. Also I thought we don't care about what happens to our predicted apogee after deploying airbrakes

airbrakes/state.py
```diff
@@ -150,21 +149,11 @@ def update(self):
...
-
         data = self.context.data_processor
 
-        # if our velocity is close to zero or negative, we are in free fall.
-        if data.vertical_velocity <= 0:
-            self.next_state()
-            return
-
-        # fallback condition:
-        # if our altitude has started to decrease, we have reached apogee:
```
harshil21: should've been commented out at least..

JacksonElia: no but its just wrong and not how we will do it

JacksonElia: Oh if you're talking about the vertical velocity thing, I just don't see any benefit of having it. Only a potential downside.

airbrakes/state.py
JacksonElia: okay I'm adding them back purely because I don't want to fix the tests lol

And its more so that it just wont work to control the air brakes doing it that way. As soon as we do that, the accel curve massively changes --> predicted apogee changes --> it's not accurate. We are going to have to write code that predicts the apogee, gets our nice dh vs velocity curve, and then does the control logic.

wlsanderson:
I know we purposely took out the airbrakes extension from the update() in coast phase, but I just added it back just so the test passes lol. We will obviously go back and change things once we do the control logic

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -93,7 +93,9 @@ def update(self, processed_data_packets: deque[ProcessedDataPacket]) -> None:
 
         :param processed_data_packets: A list of ProcessedDataPacket objects to add to the queue.
         """
-        self._prediction_queue.put(processed_data_packets)
+        # The .copy() below is critical to ensure the data is actually transferred correctly to
+        # the apogee prediction process.
+        self._prediction_queue.put(processed_data_packets.copy())
```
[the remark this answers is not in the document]
JacksonElia: does it actually make it slower???

yes that test is def a good idea. Once we get control working in another pr, we should focus on optimizing the time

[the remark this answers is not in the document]
harshil21: > does it actually make it slower???

no it was another issue

airbrakes/state.py
```diff
@@ -105,26 +105,29 @@ class MotorBurnState(State):
...
-        if data.speed < data.max_speed - data.max_speed * MAX_SPEED_THRESHOLD:
+        # We make sure that it is not just a temporary fluctuation by checking if the velocity is a
+        # bit less than the max velocity
+        if data.vertical_velocity < data.max_vertical_velocity - data.max_vertical_velocity * MAX_VELOCITY_THRESHOLD:
             self.next_state()
             return
 
         # Fallback: if our motor has burned for longer than its burn time, go to the next state
-        if time.time() - self.start_time > MOTOR_BURN_TIME:
+        if convert_to_nanoseconds(data.current_timestamp - self.start_time) > MOTOR_BURN_TIME * 1e9:
```
harshil21, line 129: no test for this

harshil21 (commented):
LGTM. Who wants the honors of merging?

JacksonElia:
I do
u gotta approve it

JacksonElia (approved):

PR #49, opened 2024-10-18
Add `FlightDisplay` + some QoL improvements

harshil21:
Notable changes:

- Use a class to handle displays
- Flight data is not lost upon Ctrl+C
- Mock sim auto terminates when flight data is exhausted
- Some more banners


main.py
```diff
@@ -22,11 +22,13 @@
     SIMULATION_LOG_PATH,
     UPSIDE_DOWN,
 )
-from utils import prepare_process_dict, update_display
+from utils import FlightDisplay
```
JacksonElia, line 25: if we're making a whole class for this, I think we should move it to `mock/`

utils.py
```diff
@@ -45,66 +53,86 @@ def deadband(input_value: float, threshold: float) -> float:
...
-    all_processes = {}
-    imu_process = airbrakes.imu._data_fetch_process
-    log_process = airbrakes.logger._log_process
-    current_process = multiprocessing.current_process()
-    for p in [imu_process, log_process, current_process]:
-        all_processes[p.name] = psutil.Process(p.pid)
-    return all_processes
+
+    # Initialize Colorama
+    init(autoreset=True)  # Automatically reset colors after each print
```
JacksonElia, line 62: shouldn't this be in the init?

JacksonElia (changes requested):
Overall looks great! Biggest change would be moving it to mock I think

JacksonElia:
I don't think you committed the change of actually moving it to mock? It isn't in mock.

harshil21:
oops

airbrakes/mock/display.py
```diff
@@ -0,0 +1,104 @@
...
+    def prepare_process_dict(self) -> dict[str, psutil.Process]:
+        """
+        Prepares a dictionary of processes to monitor CPU usage for.
+        """
+        all_processes = {}
+        imu_process = self.airbrakes.imu._data_fetch_process
+        log_process = self.airbrakes.logger._log_process
+        current_process = multiprocessing.current_process()
+        for p in [imu_process, log_process, current_process]:
+            all_processes[p.name] = psutil.Process(p.pid)
```
JacksonElia, line 103: I'm gonna nit pick you here, method needs :return in docstring and maybe add a comment for this line for psutil

harshil21: what do you want the comment to tell? Also, I love nitpicks.

JacksonElia: probably what `psutil.Process(p.pid)` does -- someone unfamiliar to what is happening might get confused

JacksonElia (approved):
LGTM

PR #51, opened 2024-10-19
`MockIMU` upgrades, better argument parsing, new `MockLogger` class

harshil21:
This PR has 3 changes:

1. New parameter for `MockIMU` to start after the log buffer. The position of the buffer is calculated dynamically.
2. Adds `MockLogger`, whose current job is to simply delete the log file after the simulation.
3. Adds more arguments to `main.py`, using the `argparse` module.

To be merged after #49


airbrakes/mock/mock_logger.py
```diff
@@ -0,0 +1,22 @@
+"""Mock Logger class for testing purposes. Currently only used to delete the log file."""
```
JacksonElia, line 1: I like this class a lot

main.py
```diff
@@ -30,13 +31,14 @@ def main(is_simulation: bool, real_servo: bool) -> None:
     sim_time_start = time.time()
 
     if is_simulation:
-        imu = MockIMU(SIMULATION_LOG_PATH, real_time_simulation=True)
+        imu = MockIMU(SIMULATION_LOG_PATH, real_time_simulation=True, start_after_log_buffer=True)
         servo = Servo(SERVO_PIN) if real_servo else Servo(SERVO_PIN, pin_factory=MockFactory(pin_class=MockPWMPin))
+        logger = MockLogger(LOGS_PATH, delete_log_file=True)
```
JacksonElia, line 36: maybe make this a command line argument, that defaults to deleting the file. make it whatever you want -- e.g. k for keep file

airbrakes/mock/mock_imu.py
```diff
@@ -19,12 +19,14 @@ class MockIMU(IMU):
...
     __slots__ = ()
 
-    def __init__(self, log_file_path: Path, real_time_simulation: bool):
+    def __init__(self, log_file_path: Path, real_time_simulation: bool, start_after_log_buffer: bool = False):
         """
         Initializes the object that pretends to be an IMU for testing purposes by reading from a log file.
         :param log_file_name: The name of the log file to read data from.
         :param real_time_simulation: Whether to simulate a real flight by sleeping for a set period, or run at full
             speed, e.g. for using it in the CI.
+        :param start_after_log_buffer: Whether to send the data packets only after the log buffer
```
JacksonElia, line 28: Is this so when testing you don't have to wait for it to launch?

airbrakes/mock/mock_imu.py
```diff
@@ -34,22 +36,59 @@ def __init__(self, log_file_path: Path, real_time_simulation: bool):
...
+                    init_timestamp = timestamp
+
         with log_file_path.open(newline="") as csvfile:
             reader = csv.DictReader(csvfile)
 
             # Reads each row as a dictionary
             row: dict[str, str]
-            for row in reader:
+            for idx, row in enumerate(reader):
+                # If we are starting after the log buffer, skip the rows before the buffer:
```
JacksonElia, line 88: Can you make it so it doesn't completely skip StandByState? Like maybe it a second before the buffer ends, or smth like that?

airbrakes/mock/mock_imu.py
harshil21: tbh @wlsanderson had problems with this for apogee prediction, that's why I did this. But it also helps quicken things up ever so slightly

airbrakes/mock/mock_logger.py
harshil21: what's there to like tho

main.py
harshil21: I think I'm going to use argparse now. I also want to add another argument on whether to run the mock sim in real time or not.

airbrakes/mock/mock_imu.py
harshil21: it does not completely skip Standbystate. The log buffer also writes 5000 packets before Motorburn, which is Standbystate

JacksonElia: Oh so it only skips the first buffer

harshil21: yep

airbrakes/mock/mock_logger.py
JacksonElia: just a simple and clean way to implement it

main.py
```diff
@@ -67,23 +68,22 @@ def main(is_simulation: bool, real_servo: bool) -> None:
...
     finally:
         airbrakes.stop()
 
 
 if __name__ == "__main__":
-    # We check if the user has passed the MOCK_ARGUMENT (m) or REAL_SERVO_ARGUMENT (r) to the program
-    # If you want to run a simulation on your computer (or Pi) run: python main.py m
-    # If you want to run a simulation on the Pi with the servo connected run: python main.py m rs
-    main(len(sys.argv) > 1 and MOCK_ARGUMENT in sys.argv[1:], len(sys.argv) > 1 and REAL_SERVO_ARGUMENT in sys.argv[1:])
+    args = arg_parser()  # Load all command line options
```
JacksonElia, line 88: I would add a comment here briefly describing each argument and what it does--it just makes it a little easier for someone new going through the code to immediately understand things

JacksonElia (approved):

PR #55, opened 2024-10-25
Run `FlightDisplay` in another thread, fix cpu usage

harshil21:
`FlightDisplay` is now on another thread, separate from main, which slightly simplifies the logic in main.py. Added benefit is that we can poll for the CPU usage separately (which is a blocking operation)

Also actually fixes the CPU usage of the processes (before it was only polling main for all the processes because of my dumb mistake) and is now normalized for number of cpu cores on your machine

See image below

image

### DIFF
--- a/airbrakes/mock/display.py
+++ b/airbrakes/mock/display.py
@@ -1,6 +1,7 @@
 """File to handle the display of real-time flight data in the terminal."""
 
 import multiprocessing
+import threading
 import time
 from typing import TYPE_CHECKING, Literal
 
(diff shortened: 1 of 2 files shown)


airbrakes/mock/display.py
```diff
@@ -40,9 +50,54 @@ def __init__(self, airbrakes: "AirbrakesContext", start_time: float) -> None:
...
+        self._thread_target = threading.Thread(target=self.update_display, daemon=True, name="Real Time Display Thread")
+        self._cpu_thread = threading.Thread(target=self.update_cpu_usage, daemon=True, name="CPU Usage Thread")
+        # Create events to signal the end of the simulation.
+        self.end_sim_natural = threading.Event()
+        self.end_sim_interrupted = threading.Event()
+
+    def start(self) -> None:
+        """Starts the display and cpu monitoring thread. Also initializes the processes to monitor."""
+        # Prepare the processes for monitoring in the simulation:
+        # This should only be done after airbrakes.start() is called, because we need the process IDs.
```
JacksonElia, line 65: move this to the docstring

JacksonElia (approved):
Love this, good job

Issue #36, opened 2024-10-03
Small issues to be fixed


harshil21:
Perfect tasks for newcomers:

- [x] Round decimal places for the fields `speed`, `avg_acceleration`, and `altitude` in `ProcessedDataPacket`.
- [x] Create an alias on the raspberry pi to run `source .venv/bin/activate` without typing out the full thing
- [x] Remove startup script on the raspberry pi (edit `/etc/rc.local`).
- [x] ~Create new Enum classes in `constants.py` like the exisiting `ServoExtension` class.~ #68 is scrapped for reasons mentioned in the PR.
- [x] Create a new mermaid diagram for the readme based on the architecture diagram below: (Taken by @Danuragtiwari)

![image](https://github.com/user-attachments/assets/39ab8ac4-83e9-4fc9-896e-1fb2bc8d91ce)

----
Comment on the issue below to get assigned!

Danuragtiwari:
Hi @harshil21, 

I’m excited to contribute to this project! Please assign this issue to me, and I’ll start working on it.

Looking forward to collaborating!

Best regards,
Anurag Tiwari

JacksonElia:
Out of curiosity, what has you interested in our code?

harshil21:
@Danuragtiwari Hi! I took a quick look at your profile and from what I can gather I think you would be well suited to do the last bullet point (mermaid diagram)

Some additional notes on that:

- The mermaid flowchart doesn't have to be exactly like I've drawn it, but it should mention that the Logger and IMU class will be handled in child processes. 

- It would be nice to use bubble shapes for the methods, ovals for classes, and squares for the output of the class (e.g. speed, avg_accel, and altitude)

- Bonus points for making it more beautiful through CSS styles.


Is it possible for you to complete this task within 2 weeks?

Danuragtiwari:
@harshil21 Thanks for considering me for this. I'll do my best to complete the task as soon as possible .

harshil21:
@Danuragtiwari alright, assigned you. Btw I'm also interested to know why you chose this repository to contribute.

Danuragtiwari:
@harshil21, @JacksonElia  I wanted to restart my open-source contributions, and I found this repository. I love problem-solving, so it excited me to join! Thanks for this opportunity :)

PR #57, opened 2024-10-26
Tiny performance improvement, type hinting improvements

harshil21:
Separates the method for setting `LoggedDataPacket`s attributes from an `IMUDataPacket`, which eliminates one extraneous `isinstance()` call, resulting in a super tiny performance gain.

Some type hinting fixes as well. Also solves a point in #36.

### DIFF
--- a/airbrakes/data_handling/logged_data_packet.py
+++ b/airbrakes/data_handling/logged_data_packet.py
@@ -2,7 +2,7 @@
 
 import msgspec
 
-from airbrakes.data_handling.imu_data_packet import EstimatedDataPacket, IMUDataPacket
+from airbrakes.data_handling.imu_data_packet import EstimatedDataPacket, RawDataPacket
 from airbrakes.data_handling.processed_data_packet import ProcessedDataPacket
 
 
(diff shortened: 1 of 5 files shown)


airbrakes/data_handling/logged_data_packet.py
```diff
@@ -21,137 +21,137 @@ class LoggedDataPacket(msgspec.Struct):
...
-    scaledGyroX: float | None = None
-    scaledGyroY: float | None = None
-    scaledGyroZ: float | None = None
-    deltaVelX: float | None = None
-    deltaVelY: float | None = None
-    deltaVelZ: float | None = None
-    deltaThetaX: float | None = None
-    deltaThetaY: float | None = None
-    deltaThetaZ: float | None = None
+    scaledAccelX: str | None = None
```
JacksonElia, line 24: why are we setting everything as a str?

airbrakes/data_handling/logged_data_packet.py
```diff
@@ -21,137 +21,137 @@ class LoggedDataPacket(msgspec.Struct):
...
-            if imu_data_packet.deltaThetaZ is not None:
-                self.deltaThetaZ = f"{imu_data_packet.deltaThetaZ:.8f}"
-
-        # Common field between the two
-        self.invalid_fields = imu_data_packet.invalid_fields
+    def set_raw_data_packet_attributes(self, data_packet: RawDataPacket) -> None:
+        """
+        Sets the attributes of the data packet corresponding to the raw data packet. Rounds the
+        float values to 8 decimal places, if the value is a float.
+        """
```
JacksonElia, line 126:
```suggestion
        :param data_packet: The raw data packet to set the attributes from.
        """
```

airbrakes/data_handling/logger.py
```diff
@@ -87,7 +87,7 @@ def log(
         self,
         state: str,
         extension: float,
-        imu_data_packets: deque[IMUDataPacket],
+        imu_data_packets: deque[EstimatedDataPacket | RawDataPacket],
```
JacksonElia, line 90: why are we doing this? doesn't this kinda defeat the point of polymorphism or whatever

JacksonElia: I guess so you can remove the extra isinstance... still feels wrong but ig its fine

airbrakes/data_handling/logged_data_packet.py
harshil21: because we have to round the values. It was always doing that anyway. Now the type hint is correct.

airbrakes/data_handling/logger.py
harshil21: > I guess so you can remove the extra isinstance... still feels wrong but ig its fine

no that's just a type hint change to make pylance happy. It's also a little more specific but I see why one wouldn't like it. I'm fine with reverting if you wanna

JacksonElia: okay I guess not a super big deal but if you remember to, revert it

JacksonElia (approved):

PR #58, opened 2024-10-26
Terminate `main.py` after we have landed and hit the buffer

### LINKED ISSUE #53: Automatically stop the main process when we have landed
We should have a mechanism to exit() the main process safely and cleanly once the rocket has landed. This has a few advantages:

- Makes sure the log buffer (the one handled by the OS) completely flushes to the file
- Saves battery power 
- Simplifies mock sim exit handling 

We should only exit() after we have hit (our) log buffer after landing, to avoid any chances of losing data.

harshil21:
Closes #53 

The interest launch data only logged 3,917 packets (equivalent to 2.61 seconds) in the `LandedState`, which means the internal OS writing buffer kicked in, which prevented writing the full 5000 packets. Luckily, we still got touchdown data. 

This PR will resolve this issue by terminating the script after we have hit the log buffer in `LandedState`. The winter launch 2023 data (december) hits this log buffer in the `LandedState` and then auto terminates. You can check this by running `main.py -m -l -f` and comparing the number of lines in LandedState in the log file generated.

So now we should get exactly 10,000 data packets in the `LandedState`.

### DIFF
--- a/airbrakes/hardware/imu.py
+++ b/airbrakes/hardware/imu.py
@@ -70,8 +70,13 @@ class IMU:
         Stops the process separate from the main process for fetching data from the IMU.
         """
         self._running.value = False
-        # Waits for the process for a maximum of 0.1 seconds to finish before stopping it
-        self._data_fetch_process.join(timeout=0.1)
+        # Fetch all packets which are not yet fetched and discard them, so main() does not get
+        # stuck waiting for the process to finish. A more technical explanation:
+        # .put() is blocking and if the queue is full, it keeps waiting for the queue to be empty, and
+        # thus the process never .joins().
+        self.get_imu_data_packets()
+        self._data_queue.close()
+        self._data_fetch_process.join()
 
     def get_imu_data_packet(self) -> IMUDataPacket | None:
         """
--- a/airbrakes/hardware/imu.py
+++ b/airbrakes/hardware/imu.py
@@ -70,8 +70,8 @@ class IMU:
         Stops the process separate from the main process for fetching data from the IMU.
         """
         self._running.value = False
-        # Waits for the process to finish before stopping it
-        self._data_fetch_process.join()
+        # Waits for the process for a maximum of 0.1 seconds to finish before stopping it
+        self._data_fetch_process.join(timeout=0.1)
 
     def get_imu_data_packet(self) -> IMUDataPacket | None:
         """
--- a/constants.py
+++ b/constants.py
@@ -73,7 +73,7 @@ STOP_SIGNAL = "STOP"
 
 
 # Don't log more than x packets for StandbyState and LandedState
-LOG_CAPACITY_AT_STANDBY = 5000  # This is equal to (x/2 + x = 3x/2 = 5000 => x = 3333 = 3.33 secs of data)
+IDLE_LOG_CAPACITY = 5000  # This is equal to (x/2 + x = 3x/2 = 5000 => x = 3333 = 3.33 secs of data)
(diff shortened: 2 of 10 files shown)


JacksonElia:
Wait can you explain your comment better? Wdym the internal OS buffer kicked in? Why would it wait to do that until landed state.

harshil21:
> Wait can you explain your comment better? Wdym the internal OS buffer kicked in? Why would it wait to do that until landed state.

From "How Linux Works, 3rd edition" by Brian Ward:

> Linux, like other Unix variants, buffers writes to the disk. This means
> the kernel usually doesn’t immediately write changes to filesystems when
> processes request changes. Instead, it stores those changes in RAM until
> the kernel determines a good time to actually write them to the disk. This
> buffering system is transparent to the user and provides a very significant
> performance gain. 

Do you remember that time a few days before interest launch when I was writing tests for the log buffer and I encountered that behavior where the number of lines reported to be cleared from the buffer just didn't match the number of lines actually written to the file? It's the same thing happening here. Unless the process / file is actually closed, the OS will keep it in RAM.

It did not "wait" do it till landed state, it's happening all the time. It's just that we stopped telling it to write at 5000 lines, but the OS kept the difference in memory, thinking that more lines are going to be added. 

So this is the sequence, in order:

1. Keep writing to the file normally until the 5000th packet in LandedState.
2. The OS tells us that it wrote till there (but as explained above, it did not, yet.)
3. We start filling up our internal log buffer, which also has a total size of 5000 packets.
4. If our buffer is full, it just keeps the latest packet, and discards the oldest packet to maintain the 5000 equilibrium.
5 a). An abrupt stop to the program (e.g. power loss) will cause us to lose our internal buffer and the OS buffer (leading to the problem of only 5000 - x packets actually written to file.
5 b). In contrast, now when we stop the program, we tell the OS to log our internal buffer (assuming it will be full) - 5000 packets. And then exit normally, so it will also flush it's own internal buffer. So we should end up with the full 10000 packets.

JacksonElia (approved):

airbrakes/hardware/imu.py
```diff
@@ -70,8 +70,8 @@ def stop(self) -> None:
         Stops the process separate from the main process for fetching data from the IMU.
         """
         self._running.value = False
-        # Waits for the process to finish before stopping it
-        self._data_fetch_process.join()
+        # Waits for the process for a maximum of 0.1 seconds to finish before stopping it
+        self._data_fetch_process.join(timeout=0.1)
```
JacksonElia, line 74: Make this a constant

constants.py
```diff
@@ -73,7 +73,7 @@ class ServoExtension(Enum):
 
 
 # Don't log more than x packets for StandbyState and LandedState
-LOG_CAPACITY_AT_STANDBY = 5000  # This is equal to (x/2 + x = 3x/2 = 5000 => x = 3333 = 3.33 secs of data)
+IDLE_LOG_CAPACITY = 5000  # This is equal to (x/2 + x = 3x/2 = 5000 => x = 3333 = 3.33 secs of data)
```
JacksonElia, line 76: I like showing the math in the comment, it has me thinking though, maybe it would be more helpful to make this a number that evenly corresponds with a certain amount of seconds? E.g. 4 seconds so 6000 packets?

Alternatively, make a different constant for amount of seconds we want the buffer to be for and then have it just do the math.

Idk just ideas, isn't really that important either way

harshil21: we can just make a function which converts the seconds to packets and then we can call it from here. I don't want another constant though. But I'm not going to do that in this PR

JacksonElia (approved):

PR #59, opened 2024-10-26
Attempted fix for landed state detection

JacksonElia:
Due to a couple of bugs, one of them being fixed in #58, we didn't actually log any data while the rocket was landed. You can see in the interest launch graph that the slope of the altitude line is still nonzero. This attempts to fix it, I will write tests for it tomorrow. 

You can see in the plot below that it doesn't actually detect landed state. This is to be expected as the launch data ends before the rocket landed.
image

I tried testing it with `winter_2023_Gyro_Comp.csv` but the mock sim almost immediately crashed without error (most likely in a different process), so that's some other problem we need to fix.


constants.py
```diff
@@ -106,8 +106,9 @@ class ServoExtension(Enum):
 
 # Free fall to Landing:
 
-# Consider the rocket to have landed if it is within 15 meters of the launch site height.
-GROUND_ALTITUDE = 15.0  # meters
+# Consider the rocket to have landed if it is within 15 meters of the launch site height and the speed is low.
+GROUND_ALTITUDE = 10.0  # meters
+LANDED_SPEED = 5.0  # m/s
```
harshil21, line 111: I almost did this as well, but then our vertical velocity still showed -15m/s ish after landing (interest launch), so I was not sure if that's a good measure. Maybe if we give it enough time the velocity would come down, it's just that the data ended prematurely so I can't be sure.

harshil21 (commented):
We should probably wait until #47 is merged and after testing with the physical IMU to check what velocity values we get (e.g. through a drop test)

constants.py
JacksonElia: It showed this because we cut off the launch data before it actually landed I think. I think we should keep something like this and just guess the values, because worst case it doesn't do the state change and we just get all the data to log.

harshil21: For the interest launch we definitely got touchdown (spike in accel). For purple nurple yeah idk it might have been cutoff

JacksonElia: Maybe then for interest launch my file I was using for the sim got cut off at the end? You can see the slope of the altitude is not zero at the end.

harshil21: I think you're using the wrong data? The graph in the readme shows LandedState.

harshil21: oh it might have been will's new data also, I haven't looked at that

JacksonElia: No my data is the same. the graph in the readme incorrectly detects the landed state, you can see it is still sloping down. Landed state should have a slope of 0.

harshil21: > incorrectly detects

yeah it intentionally switches over at 15m, so the slope won't be zero. But we can confirm that we actually got touchdown because there is a spike in the acceleration values for both the launches in the landed state. (Run the sim from my branch with `-f` and `-l` and scroll through LandedState accel values)

harshil21: But, yeah I get your point that we shouldn't switch over to LandedState arbitrarily like this. imo acceleration values should be the easiest (and safest?) way to detect landing, irrespective of height of the site. But maybe we can test with the IMU first to see what the rotated velocity values are like 1 second after landing

harshil21:
I will change the csv file 😭 so the integration test passes

airbrakes/state.py
```diff
@@ -173,8 +174,8 @@ def update(self):
 
         data = self.context.data_processor
 
-        # If our altitude is 0, we have landed:
-        if data.current_altitude <= GROUND_ALTITUDE:
+        # If our altitude and speed are around 0, we have landed
+        if data.current_altitude <= GROUND_ALTITUDE and abs(data.vertical_velocity) <= LANDED_SPEED:
```
harshil21, line 178: maybe as a fallback we can have time in free fall state should not be greater than 80 seconds?

JacksonElia: I'm okay with that

airbrakes/state.py
```diff
@@ -173,8 +175,12 @@ def update(self):
...
+            self.next_state()
+
+        # If we have been in free fall for too long, we move to the landed state
+        if convert_to_seconds(data.current_timestamp - self.start_time_ns) >= MAX_FREE_FALL_LENGTH:
```
harshil21, line 183: can we have this in combination with altitude < 15 just to be extra safe

JacksonElia: that kinda defeats the point of the fallback IMO

harshil21: ok. I was imagining a scenario where the chutes are deployed and the gust is so strong that it takes a really long time to land

JacksonElia: I talked to trent, he said the max amount of time he could ever see it being in the air (even if winds were insane) would be 180 seconds, so I'm just gonna make it that.

harshil21 (approved):

PR #60, opened 2024-10-26
added path argument to mock sim

wlsanderson:
added --path argument to main.py so that users can specify the pathname of flight data used in the mock simulation. By default, the path is scripts/imu_data/InterestLaunch-9-28.csv.


JacksonElia (commented):
I like this, it's definitely better. Do you think you could rename the flight data files to make them easier to type? Probably something along the lines of all lowercase with simpler names. 

I'm not super familiar with argparse, but could you make it so rather than having to type the entire string "scripts/imu_data/name", you just type name? 

Once you've done that, one further step would be calling lower() and trim() on the path string.

harshil21:
No I disagree with renaming the csv file. Just use <Tab> to autocomplete your path. However, I am in favour of moving the csv file to something like `launch_data/` instead of in `scripts/` which doesn't make much sense tbh.

wlsanderson:
I agree with making a launch_data folder for the flight logs. should the arg require `launch_data/name` though? or just `name`?

wlsanderson:
created launch_data folder, and changed the csv's to be lowercase and snake case. I looked into not needing to type `launch_data/`, but there would need to be quite a bit more code and other libraries added to have tab autocomplete work as if you were actually in the launch_data folder.

utils.py
```diff
@@ -82,8 +81,8 @@ def arg_parser() -> argparse.Namespace:
         "-p",
         "--path",
         help="Define the pathname of flight data to use in mock simulation. Interest Launch data is used by default",
-        type=Path,
-        default="scripts/imu_data/InterestLaunch-9-28.csv",
+        type=str,
+        default="launch_data/interest_launch_9_28.csv",
```
harshil21, line 85: why not a Path object?

launch_data/interest_launch_9_28.csv
harshil21, line 1: did you make any other changes apart from renaming and moving?

utils.py
wlsanderson: I added .lower() and .strip() to main.py, so the path argument would be filtered with those functions, then converted to a path object when put into the MockIMU class

launch_data/interest_launch_9_28.csv
wlsanderson: added the .lower() and .strip(), and also changed some references in conftest.py so they would go to the updated file path.

JacksonElia:
I didn't mean have tab auto complete work, I mean you just type `python main.py -m -p "interest_launch.csv"` rather than `python main.py -m -p "launch_data/interest_launch.csv"` or whatever

launch_data/interest_launch_9_28.csv
harshil21: that isn't what I meant, but I figured it out anyway

utils.py
wlsanderson: I had it as path object originally, but couldn't do .lower() and .strip() on path objects. I can just revert that though and keep the `type=Path`

harshil21: yeah revert them and keep it as type=Path. No need to post process either by doing .lower() and .strip().

harshil21:
>I mean you just type `python main.py -m -p "interest_launch.csv"` rather than `python main.py -m -p "launch_data/interest_launch.csv"` or whatever

no we shouldn't do dark magic to get that working. Tab autocomplete works for me though, I don't have to specify the full path. It probably depends on what shell you use. E.g. If you use zsh as a shell, one can just type `python main.py -m -p l/in` and press Tab and it will correct the path for me.

harshil21 (approved):
LGTM

PR #62, opened 2024-10-27
More tests

JacksonElia:
In this PR I:
- Cleaned up and fixed DataProcessor
- Fixed all existing tests
- Wrote tests for the gravity vector stuff and quat orientation
- Got tests for the accel rotations and vertical velocities


tests/test_data_processor.py
```diff
@@ -369,13 +370,60 @@ def test_max_altitude(self, data_processor):
...
+                    estOrientQuaternionY=0.3,
+                    estOrientQuaternionZ=0.4,
+                    estGravityVectorX=-9.6,
+                    estGravityVectorY=-0.6,
+                    estGravityVectorZ=0.5,
+                ),
+            ]
+        )
+
+        assert d._gravity_upwards_index == 0
```
harshil21, line 425: can you also test gravity_direction here

tests/test_data_processor.py
```diff
@@ -369,13 +370,60 @@ def test_max_altitude(self, data_processor):
...
+                    estOrientQuaternionY=0.3,
+                    estOrientQuaternionZ=0.4,
+                    estGravityVectorX=0.1,
+                    estGravityVectorY=-0.6,
+                    estGravityVectorZ=9.8,
+                ),
+            ]
+        )
+
+        npt.assert_array_equal(d._current_orientation_quaternions, np.array([0.1, 0.2, 0.3, 0.4]))
```
harshil21, line 402: nice, I didn't know numpy.testing existed

tests/test_data_processor.py
```diff
@@ -369,12 +478,64 @@ def test_max_altitude(self, data_processor):
         ],
     )
     def test_calculate_rotations(self, data_packets, expected_value):
```
harshil21, line 480: I want some more cases on this test. Another case should involve e.g. different axis of gravity vector, different gravity sign, no gyro data, etc.

tests/test_data_processor.py
```diff
@@ -369,12 +478,64 @@ def test_max_altitude(self, data_processor):
...
         assert len(rotations) == 3
-        for rot, expected_val in zip(rotations, expected_value, strict=False):
-            assert rot == pytest.approx(expected_val)
+        # Rotations has 3 arrays in it corresponding to x, y, and z
+        # I just could not understand how to get it to work with zip so I did this instead lol
+        assert rotations[0][-1] == pytest.approx(expected_value[0])
+        assert rotations[1][-1] == pytest.approx(expected_value[1])
+        assert rotations[2][-1] == pytest.approx(expected_value[2])
+
+    def test_initial_orientation(self):
```
harshil21, line 491: ideally this test should also be paramterized so it covers literally all 6(?) gravity cases. The two cases you chose are sufficient so this isn't necessary but still.

harshil21, line 491:
```suggestion
    def test_initial_orientation_and_gravity(self):
```

tests/test_data_processor.py
```diff
@@ -238,48 +348,7 @@ def test_first_update(self, data_processor, data_packets, init_alt, max_alt):
...
-                estLinearAccelY=y_accel[2],
-                estLinearAccelZ=z_accel[2],
-                estPressureAlt=0,
-            ),
-        ]
-        data_processor.update(new_packets)
-        second_vel = data_processor._previous_velocity
-
-        assert second_vel[0] > first_vel[0], "Previous velocity should be updated after each data update."
+            assert data.vertical_velocity == d._vertical_velocities[idx]
```
harshil21, line 351: please add more tests here, checking the new fields vertical_acceleration, and time_since_last_data_packet

harshil21 (changes requested):
good job on passing data processor tests! I would still like 2 more tests, for the `_multiply_quaternions` and `_calculate_quaternion_conjugate` functions.

airbrakes/data_handling/data_processor.py
```diff
@@ -93,56 +92,22 @@ def max_vertical_velocity(self) -> float:
...
-
-            # on the physical IMU there is a depiction of the orientation. If a negative direction is
-            # pointing to the sky, by convention, we define the gravity direction as negative. Otherwise if
-            # a positive direction is pointing to the sky, we define the gravity direction as positive.
-            # For purposes of standardizing the sign of accelerations that come out of calculate_rotated_accelerations,
-            # we define acceleration from motor burn as positve, acceleration due to drag as negative, and
-            # acceleration on the ground to be +9.8.
-            self._gravity_direction = 1 if gravity_orientation[self._gravity_upwards_index] < 0 else -1
+        if self._last_data_packet is None:
+            self._set_up()
```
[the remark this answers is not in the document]
JacksonElia: good nit

wlsanderson:
Regarding some of the unit tests with rotating acceleration vectors, I will most likely have to do another pass over the math. It most likely wont work given a flipped dataset where +Z is vertical. For the time being, if you are giving it normal compensated accelerations, gravity vectors, and quaternions that appear on the existing datasets, tests should pass and everything will work fine. I will most likely need to work on #63 a bit first so that I can fix the math with flipped datasets. I can then update the unit tests and add cases for all 6 axis if needed.

harshil21:
You mean only the rotated accelerations would  have an incorrect value? If that is the case, the unit tests should've already caught them in a way. Anyway, the unit tests should actually test every possible orientation and configuration, and we need to be sure the values are correct (cause it's hard to do the math manually). Do you know if there's an online calculator or something so we can double check?

tests/test_data_processor.py
```diff
@@ -55,112 +82,129 @@ def test_slots(self):
...
                 ),
             ]
         )
-
-        assert d._previous_velocity == pytest.approx((26.0, 25.0, 31.0))
-        assert d.vertical_velocity == pytest.approx(math.sqrt(26.0**2 + 25.0**2 + 31.0**2))
+        assert d._previous_vertical_velocity == pytest.approx(28.70443)
+        assert d.vertical_velocity == pytest.approx(28.70443)
+        assert len(d._vertical_velocities) == 3
+        # It's falling now so the max velocity should greater than the current velocity
```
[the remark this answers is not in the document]
JacksonElia: lol idk why i didn't do that I think i was just being braindead

tests/test_data_processor.py
JacksonElia: I'm gonna wait on this one for will

tests/test_data_processor.py
JacksonElia: it was a pain to write these and I figured the cases covered everything, but you're probably right. I think I'll also wait on will for this one.

wlsanderson:
No, It has to due with how the IMU would log data depending on orientation. With the rocket being -Z vertical (like interest launch) you would expect on launch pad assuming directly vertical:

- compensated acceleration: [0, 0, -9.8]
- gravity vector: [0, 0, 9.8]
- quaternion (not normalized): [1, ~0.01, ~0.01, -1]

From this, we rotate the acceleration, which gives a vector pretty close to [0, 0, -9.8]
we then flip the sign because the gravity vector is positive, so the result is [0, 0, 9.8]
this is correct. What I forgot to account for is that on an inverted flight (+Z is vertical), the initial quaternion would cause it to rotate 180 degrees.

- compensated acceleration: [0, 0, 9.8]
- gravity vector: [0, 0, 9.8]
- quaternion (not normalized): [1, ~0.01, ~0.01, 1]

When you do the same operations, I believe the acceleration would come out to [0, 0, -9.8], but we want it to be positive.
I'm not aware of any online calculators to help with this, but I'm going to use the physical IMU tomorrow to test some values and orientations and do the calculations, both in matlab and in the codebase unit tests.

JacksonElia:
@wlsanderson @harshil21 what do you want to do with this PR? I would prefer to merge this as it is, then make a new branch for fixing the math and improving the tests for them. Alternatively, we could make a branch from this fixing the math, merge it into this, then merge this into #47, but then I feel like we will have way too many branches and conflicts down the line and I don't want to deal with that.

Merging #47 with main is already going to be a headache as is

harshil21:
> I'm going to use the physical IMU tomorrow to test some values and orientations and do the calculations, both in matlab and in the codebase unit tests.

alright I get it now. Can you make sure to get 6 different data sets (csv files) for all the orientations? Simply running `main.py` with the Pi and IMU should give you the all the data we need. We can then easily pass a couple of those packets to the unit tests to make sure we get everything right. (perhaps in a new folder in `tests/imu_data/`?)

harshil21:
> @wlsanderson @harshil21 what do you want to do with this PR? I would prefer to merge this as it is, then make a new branch for fixing the math and improving the tests for them. Alternatively, we could make a branch from this fixing the math, merge it into this, then merge this into #47, but then I feel like we will have way too many branches and conflicts down the line and I don't want to deal with that.

Yeah a separate PR after merging this is the way to go, to keep things atomic. 

> Merging #47 with main is already going to be a headache as is

dw, there wouldn't be too many conflicts (I already solved most yesterday)

harshil21 (changes requested):
The mock sim is failing. If that passes, this PR should be good to go.

harshil21 (approved):

PR #67, opened 2024-10-29
Apogee pred tests

JacksonElia:
This pr aims to fix existing tests and add new tests related to apogee prediction

### DIFF
--- a/airbrakes/data_handling/apogee_predictor.py
+++ b/airbrakes/data_handling/apogee_predictor.py
@@ -33,11 +33,8 @@ class ApogeePredictor:
         "_cumulative_time_differences",
         "_current_altitude",
         "_current_velocity",
-        "_params",
         "_prediction_process",
         "_prediction_queue",
-        "_running",
-        "_start_time",
         "_time_differences",
     )
 
@@ -70,7 +67,7 @@ class ApogeePredictor:
         """
         Returns whether the prediction process is running.
         """
-        return self._running.value
+        return self._prediction_process.is_alive()
 
     def start(self) -> None:
         """
--- a/airbrakes/data_handling/apogee_predictor.py
+++ b/airbrakes/data_handling/apogee_predictor.py
@@ -77,7 +77,7 @@ class ApogeePredictor:
 
     def stop(self) -> None:
         """
-        Stops the logging process. It will finish logging the current message and then stop.
+        Stops the prediction process.
         """
         # Waits for the process to finish before stopping it
         self._prediction_queue.put(STOP_SIGNAL)  # Put the stop signal in the queue
@@ -162,7 +162,6 @@ class ApogeePredictor:
         signal.signal(signal.SIGINT, signal.SIG_IGN)  # Ignores the interrupt signal
         # These arrays belong to the prediction process, and are used to store the data packets that are passed in
         last_run_length = 0
-
         # Keep checking for new data packets until the stop signal is received:
         while True:
             # Rather than having the queue store all the data packets, it is only used to communicate between the main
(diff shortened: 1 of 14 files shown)


airbrakes/data_handling/apogee_predictor.py
```diff
@@ -152,10 +149,15 @@ def _get_apogee(self, params: npt.NDArray[np.float64]) -> np.float64:
         if params is None:
             return 0.0
```
JacksonElia, line 150: why do we return 0? I know its like convention or whatever to return 0 to show it ended without problems. At the very least if we are returning something, add it to the docstring and the return type

constants.py
```diff
@@ -112,3 +112,4 @@ class ServoExtension(Enum):
 # The altitude at which the rocket is expected to reach apogee, without the airbrakes
 TARGET_ALTITUDE = 1700  # m (5,100 ft)
 CURVE_FIT_INITIAL = [-10.5, 0.03]
+APOGEE_PREDICTION_FREQUENCY = 100  # packets
```
JacksonElia, line 115: this can and should me MUCH lower imo

airbrakes/data_handling/apogee_predictor.py
harshil21: I don't know if there ever is a case where we did return 0.0. The coverage would help verify that actually. So i'll wait for that to be merged.

constants.py
JacksonElia: when we looked at CPU  usage, it was like 3% with this. If we lower this to 2 packets, CPU usage still is only ~20%. This change could get us a prediction like .1 or .2 seconds faster, and we really need every tenth of a second we can get. Ideally we can lower the prediction time to ~1.5 seconds.

airbrakes/data_handling/apogee_predictor.py
JacksonElia: wait disregard this one, I'm tweaking.

the :return in docstring should probably still specify it returns 0.0 if there was a problem/no apogee predicted

constants.py
harshil21: well I'd be down to lower it to 50 packets, but anything lower is just useless compute imo

wlsanderson:
Added multiprocessing events, so the unit tests can wait until apogee prediction is finished to proceed. Due to how windows handles multiprocessing, it can sometimes take a few seconds for the first update to run, compared to milliseconds on linux.

constants.py
JacksonElia: but like why not do it faster? Thats still a .1 second delay. Like I would get it if we needed the extra performance, but we don't--we're not even close to hitting it. 

The faster we computer it, the more frequently/faster we can check it converges.

wlsanderson: I'd lower it to around 10 data packets maybe. Even though the speed of convergence is pretty much negligible, it would be good to log it, and more data points gives better post-processing

harshil21: alrightttt i'll cave

JacksonElia:
Once the constant is lowered to 10 packets, I'll merge

harshil21:
No I'm fixing the integration test

tests/test_apogee_predictor.py
```diff
@@ -0,0 +1,160 @@
...
+        ap._prediction_complete.wait(timeout=5)
+
+        predicted_apogee = ap.apogee
+        # Verify result is a float and equal to predicted value
+        assert isinstance(predicted_apogee, float)
+        assert predicted_apogee == pytest.approx(expected_value)
+        ap.stop()
+
+    def test_prediction_loop_every_x_packets(self, apogee_predictor):
+        """Tests that the predictor only runs every 100 packets"""
```
harshil21, line 129:
```suggestion
        """Tests that the predictor only runs every APOGEE_PREDICTION_FREQUENCY packets"""
```

harshil21 (approved):
one tiny change

PR #68, opened 2024-10-30
All Constants as Enum

wlsanderson:
Branch is taken from apogee-pred, so some tests still fail, but Harshil said it looked good and none of the fails were due to the changes.

ToDo: Integration test. This enum change is required to write a good integration test.

### DIFF
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -61,7 +61,7 @@ class AirbrakesContext:
         self.apogee_predictor = apogee_predictor
 
         # Placeholder for the current airbrake extension until they are set
-        self.current_extension: ServoExtension = ServoExtension.MIN_EXTENSION
+        self.current_extension: ServoExtension = ServoExtension.MIN_EXTENSION.value
         # The rocket starts in the StandByState
         self.state: State = StandByState(self)
         self.shutdown_requested = False
@@ -142,11 +142,11 @@ class AirbrakesContext:
         Extends the airbrakes to the maximum extension.
         """
         self.servo.set_extended()
-        self.current_extension = ServoExtension.MAX_EXTENSION
+        self.current_extension = ServoExtension.MAX_EXTENSION.value
 
     def retract_airbrakes(self) -> None:
         """
         Retracts the airbrakes to the minimum extension.
         """
         self.servo.set_retracted()
-        self.current_extension = ServoExtension.MIN_EXTENSION
+        self.current_extension = ServoExtension.MIN_EXTENSION.value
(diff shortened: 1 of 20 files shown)


airbrakes/airbrakes.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         self.apogee_predictor = apogee_predictor
 
         # Placeholder for the current airbrake extension until they are set
-        self.current_extension: ServoExtension = ServoExtension.MIN_EXTENSION
+        self.current_extension: ServoExtension = ServoExtension.MIN_EXTENSION.value
```
JacksonElia, line 64: We should be leaving this as the enum rather than getting its value. If for some reason you want extensions to be represented as floats rather than as enums that you unpack, then we need to change all of the types back to floats.

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -83,7 +83,7 @@ def stop(self) -> None:
         Stops the logging process. It will finish logging the current message and then stop.
         """
         # Waits for the process to finish before stopping it
-        self._prediction_queue.put(STOP_SIGNAL)  # Put the stop signal in the queue
+        self._prediction_queue.put(LoggerSettings.STOP_SIGNAL.value)  # Put the stop signal in the queue
```
JacksonElia, line 86: You do not need an enum for a single value, just use a constant

JacksonElia: like if we had 3 or 4 different signals we used, then I could see an argument for it

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -117,7 +117,7 @@ def _curve_fit(self) -> npt.NDArray[np.float64]:
             self._curve_fit_function,
             self._cumulative_time_differences,
             self._accelerations,
-            p0=CURVE_FIT_INITIAL,
+            p0=ApogeePredictorSettings.CURVE_FIT_INITIAL.value,
```
JacksonElia, line 120: Similar to my previous comment, this should not be an enum. This kinda misinterprets the idea/use case of enums. 

They should be used when you have a place that can only accept a set of specific values. E.g. you could have a compass class that currently represents the direction it is facing with an enum for direction containing the values North, South, East, and West.

In this case, it doesn't make sense to use an enum as 1. we only have two values, but 2. (and more importantly) the place that this is being used will only ever use this one specific constant. It won't ever use other values from the apogee predictor settings.

JacksonElia:
I disagree with this PR as I feel it fundamentally misses what the point of enums are for. Maybe I'm wrong and python just has a different convention than normal. IDK

harshil21:
Yeah you'd be right. The problem is we have too many mixed types and there is no way to group one of the types together under one kind of category. The `IntEnum` wouldn't even work for floats and just converts them to an `int`.

Well at least now we know.

Sorry @wlsanderson, but I will cherry-pick https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/68/commits/0b858a9eb2123dd50c083e36d080e5d2cacb815e into apogee-pred-tests at least.

PR #69, opened 2024-10-31
Use `scipy` for rotations, improve performance of time differences by ~20%

harshil21:
- Uses scipy for calculating rotations. This is faster, and less code!
- Tiny change in `_calculate_time_differences` leads to a roughly 19% performance improvement. 

(Credit to AI for helping me discover that scipy has a built in function to do rotations in quaternions!)

Sorry @JacksonElia some of your data processor tests aren't needed anymore, that's just the way it goes

----
Performance testing script:

```python
import timeit

time_diff_1 = """
import numpy as np
import msgspec
def _calculate_time_differences(input_vec):
    return np.diff([data_packet.timestamp for data_packet in input_vec]) * 1e-9

class A(msgspec.Struct):
    timestamp: int

vec = [A(1e9), A(1e9 + 0.2), A(1e9 + 0.4), A(1e9 + 0.6), A(1e9 + 0.8)]

"""

time_diff_2 = """
import numpy as np
import msgspec
def _calculate_time_differences(input_vec):
    return np.diff([data_packet.timestamp * 1e-9 for data_packet in input_vec])

class A(msgspec.Struct):
    timestamp: int

vec = [A(1e9), A(1e9 + 0.2), A(1e9 + 0.4), A(1e9 + 0.6), A(1e9 + 0.8)]

"""
time_diff_1_time = timeit.timeit(
    stmt="_calculate_time_differences(vec)",
    setup=time_diff_1,
)


time_diff_2_time = timeit.timeit(
    stmt="_calculate_time_differences(vec)",
    setup=time_diff_2,
)
```
```
Time diff 1 time: 4.220219339003961
Time diff 2 time: 3.418315943999914
```

### DIFF
--- a/airbrakes/data_handling/data_processor.py
+++ b/airbrakes/data_handling/data_processor.py
@@ -220,7 +220,7 @@ class IMUDataProcessor:
 
             # Rotate the acceleration vector using the updated orientation
             rotated_accel = current_orientation.apply([x_accel, y_accel, z_accel])
-            # Vertical acceleration will always be the 4th element of the quaternion, regardless of orientation.
+            # Vertical acceleration will always be the 3rd element of the rotated vector, regardless of orientation.
             # For simplicity, we multiply by -1 so that acceleration during motor burn is positive, and
             # acceleration due to drag force during coast phase is negative.
             rotated_accelerations[idx] = -rotated_accel[2]
--- a/airbrakes/data_handling/data_processor.py
+++ b/airbrakes/data_handling/data_processor.py
@@ -244,6 +244,8 @@ class IMUDataProcessor:
                 for vertical_acceleration in self._rotated_accelerations
             ]
         )
+        # Technical notes: Trying to vectorize the deadband function via np.vectorize() or np.frompyfunc() is
+        # slower than this approach.
 
         # Integrate the accelerations to get the velocities
         vertical_velocities = self._previous_vertical_velocity + np.cumsum(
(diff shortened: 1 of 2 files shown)


harshil21:
(Tests are actually passing but since it fails on `apogee-pred` it shows failing here)

airbrakes/data_handling/data_processor.py
```diff
@@ -298,4 +267,4 @@ def _calculate_time_differences(self) -> npt.NDArray[np.float64]:
         # We are converting from ns to s, since we don't want to have a velocity in m/ns^2
         # We are using the last data point to calculate the time difference  between the last data point from the
         # previous loop, and the first data point from the current loop
-        return np.diff([data_packet.timestamp for data_packet in [self._last_data_packet, *self._data_packets]]) * 1e-9
+        return np.diff([data_packet.timestamp * 1e-9 for data_packet in [self._last_data_packet, *self._data_packets]])
```
JacksonElia, line 270: we should change this to using convertNanoseconds or whatever in utils

JacksonElia:
Ah well, that's the way it goes.

I don't really feel like trying to understand this math rn, but overall it looks good to me if the math all works out. @wlsanderson  once you say it looks good, I'm good with merging.

airbrakes/data_handling/data_processor.py
harshil21: i'm confused, we want it in seconds... not nanoseconds

wlsanderson:
I got in bed just as I saw this lol so I will have to check tomorrow and pull up the matlab to compare values. At first glance it looks good.

wlsanderson:
Checked against sets of values in both launches, and unit tests pass, so everything looks good to me

wlsanderson (approved):

PR #71, opened 2024-11-01
Adding Air Brakes Control

JacksonElia:
Working on getting control logic in place for the air brakes
- [x] Need code for convergence
- [x] ApogeePredictor needs to be cleaned up  -- better names, comments, and splitting the loop into more methods
- [x] Need code for converting predicted apogee into vel vs delta position function (or lookup table idk)


wlsanderson:
values that I think will work for subscale launch:
MAX_VELOCITY_THRESHOLD = 0.08
duration of coast phase until we stop curve fitting: 1.75 sec

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -188,12 +213,38 @@ def _prediction_loop(self) -> None:
...
+
+        # Because we return 0.0 if apogee hasn't been predicted, we don't want to say it's
+        # converged if there's any 0s
+        if np.any(self._predicted_apogees == 0):
+            return False
+
+        avg_apogee = np.mean(self._predicted_apogees[-NUMBER_OF_PREDICTIONS:])
+
+        # Check if each apogee is within 3% of the average
+        return bool(
```
harshil21, line 248: it's already a bool, no need of bool()

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -188,12 +213,38 @@ def _prediction_loop(self) -> None:
...
+                    lookup_table = self._generate_prediction(curve_coefficients)
                 self._cumulative_time_differences = np.cumsum(self._time_differences)
-                params = self._curve_fit()
-                self._apogee_prediction_value.value = self._get_apogee(params)
+                # Predicts the apogee using the lookup table and linear interpolation
+                predicted_apogee = np.interp(
+                    self._current_velocity, lookup_table[0], lookup_table[1]
+                )
+                self._apogee_prediction_value.value = predicted_apogee
+                self._predicted_apogees = np.append(self._predicted_apogees, predicted_apogee)
```
harshil21, line 227: Is there a reason we are keeping an unbounded list of predicted apogees in the first place? Why not just a deque of say, last 20 predicted apogees?

harshil21 (commented):
initial review

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -188,12 +213,38 @@ def _prediction_loop(self) -> None:
...
                 last_run_length = len(self._accelerations)
             # notifies tests that prediction is complete
             self._prediction_complete.set()
+
+    def _has_apogee_converged(self) -> bool:
+        """
+        Checks if the last 5 apogees are within 3% of each other.
+        :return: True if the last 5 apogees are within 3% of each other, otherwise False.
+        """
+        if len(self._predicted_apogees) < NUMBER_OF_PREDICTIONS:
```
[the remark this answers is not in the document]
wlsanderson: I agree with doing it based off time. Would probably need another constant, unless we are sure that 0.5 seconds is a good reference.

airbrakes/data_handling/apogee_predictor.py
JacksonElia: lol thats what I thought but pycharm was mad that it was a np.bool or some shit

airbrakes/data_handling/apogee_predictor.py
JacksonElia: i was thinking of doing that but then I would have to write all the code to like push and pop stuff and I didn't feel like doing it yet

harshil21: with deque's you don't have to manage that, it just automatically pops the last element when it's full

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -188,12 +213,38 @@ def _prediction_loop(self) -> None:
...
+                predicted_apogee = np.interp(
+                    self._current_velocity, lookup_table[0], lookup_table[1]
+                )
+                self._apogee_prediction_value.value = predicted_apogee
+                self._predicted_apogees = np.append(self._predicted_apogees, predicted_apogee)
                 last_run_length = len(self._accelerations)
             # notifies tests that prediction is complete
             self._prediction_complete.set()
+
+    def _has_apogee_converged(self) -> bool:
```
harshil21, line 232: ok I just had a deeper look on this. We should be checking apogee convergence not like this, but by using the covariance matrix returned by the curve_fit function. 

Currently with this approach, the code says that apogee has converged below 1800m for interest launch, which is wrong. A simple check like this is better imo (less code, faster):

```python
popt, pcov = scipy.optimize.curve_fit(func, xdata, ydata)
uncertainties = np.sqrt(np.diag(pcov))

if np.all(uncertainties < threshold):
    print("Fit has converged")
```

I found a good threshold to be: 
```
threshold = np.array([[0.0259, 0.00065]])
```
which works for both interest and purple launch.

harshil21 (commented):
our method for checking for convergence should be changed:

airbrakes/data_handling/apogee_predictor.py
harshil21: another alternative would be to use the standard deviation, however this is not without limitations also, since it is susceptible to local maximas of our predicted apogee graph, so it can converge early. If you use these settings:

Threshold: <= 1.7, `NUMBER_OF_PREDICITIONS = 10`, then:

converged at 1868m for int launch, at alt = 614m
converged at 795m for purple launch, at alt = 496m

airbrakes/data_handling/apogee_predictor.py
JacksonElia: lol thats what i get for asking chat--I asked it what would happen and it said it like wouldn't add 💀

airbrakes/data_handling/apogee_predictor.py
JacksonElia: okay we can add this but you'll have to explain to me how tf it works

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -97,87 +113,110 @@ def update(self, processed_data_packets: deque[ProcessedDataPacket]) -> None:
...
             self._curve_fit_function,
             self._cumulative_time_differences,
             self._accelerations,
             p0=CURVE_FIT_INITIAL,
             maxfev=2000,
-        )[0]
+        )
+        uncertainties = np.sqrt(np.diag(pcov))
+        # determines the minimum amount of data points before we can declare apogee converged or not
+        min_length = np.ceil(MIN_PREDICTION_TIME * APOGEE_PREDICTION_FREQUENCY)
```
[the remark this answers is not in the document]
wlsanderson: this is so we don't declare our apogee as converged before we reach a time (0.5 sec)

wlsanderson:
@harshil21 could you look over the monkeypatch tests? I don't really think I did them the proper way, because it mostly just felt like I was just remaking the prediction loop method lol. I'm pretty sure I had to use monkeypatch for that method though? not sure what other way I could have done it.

harshil21:
yeah I was just having a glance over them. The point of monkeypatching at least in our case is to check if certain methods are called as expected, not to duplicate the code from main code (because then the test would be detached from our library, and doesn't actually test it).

Regarding the best way to do it - I'm not sure yet. Maybe we might have to mock the multiprocessing module so it doesn't actually spawn a new process and it's slightly easier to test that way. I will take another look at these tests tonight.

The other tests look fantastic at first glance though, they touch on keys areas of the code 👍🏽

airbrakes/data_handling/apogee_predictor.py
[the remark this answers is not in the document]
harshil21: Yeah but why, isn't the standard deviation of the errors (`uncertainites`) enough already? Is there a particular case / scenario where you think that would fail?

In any case, this should be a class/module level constant, to avoid recomputation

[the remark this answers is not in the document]
wlsanderson: ah yeah you're right. I reverted that change and fixed the corresponding tests.

PR #72, opened 2024-11-01
Add ruff as pre-commit hook

### LINKED ISSUE #70: Add ruff as pre-commit check
We should add ruff as a pre-commit check, which will run on every `git commit`, so we don't have all these extraneous commits fixing ruff

harshil21:
Closes #70 

Make sure to install `pre-commit`!

`pip install pre-commit`
`pre-commit install` (inside the airbrakes folder)


JacksonElia (approved):
Yeah this is good, we used it at my internship

PR #76, opened 2024-11-05
Generated Data Simulator

wlsanderson:
Adds a new --sim argument that will run the mock simulation with randomly generated data. Long-term WIP.


simulator/sim_imu.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self._data_generator = DataGenerator()
+
+    def _fetch_data_loop(self) -> None:
+        """A wrapper function to suppress KeyboardInterrupt exceptions when obtaining generated
+        data."""
+        # unfortunately, doing the signal handling isn't always reliable, so we need to wrap the
+        # function in a context manager to suppress the KeyboardInterrupt
+        with contextlib.suppress(KeyboardInterrupt):
+            for _ in range(21000):
```
JacksonElia, line 57: Make 21000 a constant

simulator/sim_imu.py
```diff
@@ -0,0 +1,68 @@
...
+                packets = self._data_generator.generate_data_packet()
+                timestamp = 0
+                if packets[0] is not None:
```
JacksonElia, line 60: Could we instead make it so `data_generator` has methods for generating a raw data packet and generating an estimated data packet? Then here in this loop we could keep track of which data packets it should be.

simulator/data_gen.py
```diff
@@ -0,0 +1,332 @@
...
+        "_config",
+        "_current_timestamp",
+        "_last_est_packet",
+        "_last_raw_packet",
+        "_last_velocity",
+        "_thrust_data",
+        "_vertical_index",
+    )
+
+    def __init__(self):
```
JacksonElia, line 35: it would be nice if we could specify the frequency that sim_imu runs at/this generates data at. I don't think it would be that much work to implement and it will be a lot easier to do it now rather than later.

wlsanderson: not sure what you mean by this. Do you mean like, real-time simulation, vs going full speed? Or do you mean the frequency between raw and estimated data packets? If the latter, then I have constants for that in the yaml config, and those are being used to determine the frequency between data packets.

JacksonElia:
Once you are done with this, I'll look over and maybe do a little more clean up, and maybe I'll have some suggestions/changes. Then we can merge it.

Once everything is merged, we can make a new branch/pr for writing integration tests that use the sim and test air brakes working

wlsanderson:
sounds good. Im not too sure when i'll be done with it, because it mostly comes down to fixing the last few bugs with apogee convergence time, and im not sure how long it will take me to find (and fix) the bug lol

utils.py
```diff
@@ -99,17 +99,39 @@ def arg_parser() -> argparse.Namespace:
         default=False,
     )
 
+    parser.add_argument(
+        "-s",
+        "--sim",
+        help="Runs the data simulation alongside the mock simulation, with an optional scale",
```
harshil21, line 105: Maybe down the line we might have to make this a subcommand so we can pass in more arguments to the sim.
So something like `python main.py sim --type full-scale --temperature 278`

JacksonElia: atp just change the config imo

README.md
```diff
@@ -232,6 +234,10 @@ If you want to run a mock launch, but with the real servo running, run:
 ```bash
 uv run mock -r
 ```
+To run a mock launch with our custom simulation:
+```bash
+python3 main.py -m -s
```
harshil21, line 239: need to use uv. `uv run sim` or something

harshil21 (commented):
will do full review a bit later today

README.md
wlsanderson: I didn't use sim because of how `uv run mock -s` currently works. By default, `uv run mock -s` will run full scale config, but with `uv run -s sub-scale`, it will run the sub scale config. There wasn't a clean way to do this with `uv run sim` without creating another `args` for sim config type, like `uv run sim --config sub-scale`. This just seems like a worse option compared to `uv run mock -s sub-scale`

harshil21: Yeah we did discuss this. Turns out I apparently mentioned this really early on https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/76#discussion_r1841399639 but forgot about it later lol. I suppose we can write a conditional statement to change the kind of argument in the `arg_parser`. 

so something like:

```python
# in arg_parser()

if sim_invocation and not args.type:
     args.type = "full-scale"
```
so then the behaviour `uv run sim` would run full-scale by default, and `uv run sim -t sub-scale` will use sub-scale config. Thus we don't need an extra argument to pass. Unless I'm missing something

wlsanderson: Is it possible to have it be something like `uv run sim` for a default full-scale, but to specify pre-set config, it could just simply be `uv run sim sub-scale`?

harshil21: yeah. E.g:

```python
parser.add_argument("type", default="full-scale", nargs="?")
>>> parser.parse_args([])
Namespace(type='full-scale')
>>> parser.parse_args(["sub-scale"])
Namespace(type='sub-scale')
```

harshil21 (approved):

PR #77, opened 2024-11-08
Readme Additions

JacksonElia:
Just added stuff for simulations, but are there anywhere else we need to add to the readme?


wlsanderson:
what about scripts?

JacksonElia:
good idea, we can fix scripts with this

PR #79, opened 2024-11-09
Retract airbrakes when under target apogee, remove no buzz condition

JacksonElia:
@harshil21 pointed out that we didn't actually have code for retracting the air brakes after they were deployed, so this add it.


airbrakes/state.py
```diff
@@ -144,6 +144,9 @@ def update(self):
         # if our prediction is overshooting our target altitude, extend the airbrakes
         if self.context.apogee_predictor.apogee > TARGET_ALTITUDE:
             self.context.extend_airbrakes()
+        else:
+            # If our prediction is not overshooting, retract the airbrakes
+            self.context.retract_airbrakes()
```
harshil21, line 149: we need code here to not keep sending the command to the servo, like we had for interest launch. I don't know if that's a concern with the way the new servo is setup or if it's fine now and that wouldn't matter

airbrakes/state.py
```diff
@@ -144,6 +144,9 @@ def update(self):
         # if our prediction is overshooting our target altitude, extend the airbrakes
         if self.context.apogee_predictor.apogee > TARGET_ALTITUDE:
             self.context.extend_airbrakes()
+        else:
```
harshil21, line 147: for the NASA launch we have to stay within a range of altitudes correct? e.g. 6000-6500ft?

airbrakes/state.py
JacksonElia: Generally with robotics/motor control, it is fine to keep sending the command. We do have the shitty threading stuff which in that case we don't want to, but for this launch we are gonna get rid of the no buzz I think.

harshil21: alright, then I would like us to get rid of that threading stuff before launch in another PR.

airbrakes/state.py
JacksonElia: yes, from 4000-6000 ft, but we are trying to hit a specific target apogee inside of that range.

harshil21: yeah, that's what I was going to suggest. Our variable should be set maybe 60% in the range, so something like 5200ft.

JacksonElia: We have a specific height chosen by aubri. For the competition rn, we are thinking of 4800 ft

airbrakes/state.py
JacksonElia: I'm just gonna do it in this pr

harshil21 (approved):

PR #83, opened 2024-11-12
Add `--verbose` and print display on `StandbyState` for main

harshil21:


JacksonElia (approved):
@harshil21 read over this but I think its good cause I wrote it lol

harshil21 (commented):
LGTM. One thing we overlooked is that we need to be able to "quit" the process because we need to detach from our tmux session. Fortunately in Linux it is trivial to simply suspend the process once we've looked at the values, and resume the process in the background, so we can detach from the tmux session.

JacksonElia:
I'm going to go ahead and merge it, then we can check that everything works as expected tmrw

PR #100, opened 2024-11-21
Adding more fields to log

JacksonElia:
Why not log even more data? With this PR we need to log:
- [x] Make it correctly log extension
- [x] Log Convergence Parameters
- [x] Log Queue sizes
- [x] Log the first timestamp/packet which airbrakes.update() runs with

Must be merged after #99


harshil21:
I would argue that after #99 we don't need to log queue sizes anymore.

wlsanderson:
I still think we should tbh. Theres no reason not to. It definitely helps with debugging and finding where performance issues happen, and it's one of the fields where you cant go back in post-processing and find it out then.

JacksonElia:
I agree with will. If for some reason in the future they do exceed 15, we wouldn't know that easily,

wlsanderson:
i mean whats one more column on the log file. Gotta use the whole 128gb if we have 128gb

harshil21:
alright, it could be ever so slightly more overhead but it's probably worth it

harshil21:
Changes:

- All packets are now in a separate folder
- `DebugPacket` is only logged for the first packet processed, so that way we know which is the first packet the batch was, and also saves some space since those fields would be repeated for that batch anyway.

wlsanderson (commented):
I feel like the debug packet is just not an elegant solution to this issue. While I can see the purpose of logging at the start of each set of packets, this seems like it adds a lot more complexity to the issue. Passing the DebugPacket into a LoggedDataPacket just doesn't really sit right with me lol. I think just having these values logged every line (even if they don't update) is a cleaner solution, and arguably easier to do post-processing with.

harshil21:
Well we do have a lot of fields to log and the `log()` method is just getting too crowded. We are probably going to have more fields we would want to log in the future.

If we log these for the entire batch, then we would need another field to know which packet was the first one in the batch. 

I guess it's now up to @JacksonElia to break the tie.

JacksonElia:
I agree with Will here, I don't think I like the DebugDataPacket. Though because we have multiple fields we are logging from the apogee predictor, I would be fine with an apogee_prediction_data_packets. As for the queue sizes, I think they should be logged separately/individually.

After thinking about this a lot though, I think that the cleanest solution could eventually be using **kwargs when logging. Maybe I'll add it eventually.

harshil21:
> I would be fine with an apogee_prediction_data_packets.

So another data packet class but would only contain things from that process? I can get behind that. How do you want to send the packets from that process to the main process? Do you want to include `predicted_apogee` in that too? Another Queue?

> After thinking about this a lot though, I think that the cleanest solution could eventually be using **kwargs when logging. 

Where? Some more details would be nice

JacksonElia:
Me and will talked about it a lot, we agree with the need of some data packet like debug data packet.

We want to emphasize categorizing information with data packets, and paying attention to what "owns" a data packet, so we're gonna make a data packet for apogee predictor and for the context (this will be like your debug data packet). We also will probably make a servo data packet because we're gonna have the encoder info.

JacksonElia:
Okay I think now all is left is actually logging the ContextPacket and the ApogeePredictorPackets
@harshil21 if you want to do it, go for it, if not I'll work on it soon. Lmk what you think of my implementation with the queue stuff

airbrakes/airbrakes.py
```diff
@@ -74,7 +80,9 @@ def __init__(
         self.shutdown_requested = False
         self.imu_data_packets: deque[IMUDataPacket] = deque()
         self.processed_data_packets: list[ProcessedDataPacket] = []
+        self.apogee_predictor_data_packets: list[ApogeePredictorDataPacket] = []
         self.est_data_packets: list[EstimatedDataPacket] = []
+        self.debug_packet: ContextPacket | None = None
```
wlsanderson, line 85: do we plan to just keep it named debug_packet?

wlsanderson (commented):
We should probably include state in the Context Packet, as well as either making a servo packet (with extension and eventually encoder) or just put servo extension into Context Packet as well

airbrakes/airbrakes.py
JacksonElia: no that will be changed

JacksonElia:
Yes state should be in ContextPacket and we can make a ServoPacket

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -128,17 +141,26 @@ def stop(self) -> None:
...
+        self._processed_data_packet_queue.put_many(processed_data_packets)
+
+    def get_prediction_data_packets(self) -> list[ApogeePredictorDataPacket]:
+        """
+        Gets the apogee prediction data packets from the queue.
+        """
+        return [
+            self._apogee_predictor_packet_queue.get()
+            for _ in range(self._apogee_predictor_packet_queue.qsize())
+        ]
```
harshil21, line 163: use `queue.get_many()`

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -249,17 +278,35 @@ def _prediction_loop(self) -> None:
...
             if len(self._accelerations) - last_run_length >= APOGEE_PREDICTION_MIN_PACKETS:
                 self._cumulative_time_differences = np.cumsum(self._time_differences)
-                # We only want to keep curve fitting if the curve fit hasn't converged yet
-                if not self._has_apogee_converged:
-                    curve_coefficients = self._curve_fit()
-                    self._update_prediction_lookup_table(curve_coefficients)
+
+                # Placeholder values for the curve coefficients and apogee
+                curve_coefficients = -1, -1
+                apogee = -1
```
harshil21, line 284: let's use `self.apogee` and `self.curve_coefficients` and initialize them to `None`

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -249,17 +278,35 @@ def _prediction_loop(self) -> None:
...
 
                 # Get the predicted apogee if the curve fit has converged:
                 if self._has_apogee_converged:
-                    self._predict_apogee()
+                    apogee = self._predict_apogee()
+                else:
+                    # We only want to keep curve fitting if the curve fit hasn't converged yet
+                    curve_coefficients = self._curve_fit()
+                    self._send_to_main_process(curve_coefficients)
```
harshil21, line 292:
```suggestion
```
bye

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -249,17 +278,35 @@ def _prediction_loop(self) -> None:
...
-    def _extract_processed_data_packets(self, data_packets: list[ProcessedDataPacket]) -> None:
+                # Send the apogee prediction to the main process
+                self._apogee_predictor_packet_queue.put(
+                    ApogeePredictorDataPacket(
+                        predicted_apogee=apogee,
+                        prediction_has_converged=self._has_apogee_converged,
+                        a_coefficient=float(curve_coefficients.A),
+                        b_coefficient=float(curve_coefficients.B),
+                        uncertainty_threshold_1=self.uncertainty_threshold_1.value,
+                        uncertainty_threshold_2=self.uncertainty_threshold_2.value,
```
harshil21, line 305:
```suggestion
                        uncertainty_threshold_1=self.uncertainty_threshold_1,
                        uncertainty_threshold_2=self.uncertainty_threshold_2,
```

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -103,19 +117,18 @@ def __init__(self):
...
-        Returns the predicted apogee of the rocket
-        :return: predicted apogee as a float.
+        Returns whether the prediction process is running.
         """
-        return float(self._apogee_prediction_value.value)
+        return self._prediction_process.is_alive()
 
     @property
-    def is_running(self) -> bool:
+    def queue_size(self) -> int:
```
harshil21, line 127: name is ambiguous since we now have 2 queues

airbrakes/airbrakes.py
```diff
@@ -129,31 +138,53 @@ def update(self) -> None:
         if self.est_data_packets:
             self.processed_data_packets = self.data_processor.get_processed_data_packets()
 
+        # Gets the apogee predictor packets
+        apogee_predictor_data_packets = self.apogee_predictor.get_prediction_data_packets()
+        if apogee_predictor_data_packets:
+            # We only want to update the apogee predictor data packets if we have new data
+            self.apogee_predictor_data_packets = apogee_predictor_data_packets
```
harshil21, line 145: let's assign `self.apogee..` to `None` when no packets were fetched, so we only get them for CoastState

airbrakes/data_handling/logger.py
```diff
@@ -123,55 +126,62 @@ def _convert_unknown_type(obj_type: Any) -> str:
...
-                # Add the predicted apogee field:
-                logged_fields["predicted_apogee"] = predicted_apogee
                 index += 1
 
-            logged_data_packets.append(logged_fields)
+            # Apogee Prediction happens asynchronously, so we need to check if we have a packet
+            # This means that rather than belonging to a specific IMU packet, it belongs to the
+            # context packet
+            if apogee_predictor_data_packets:
+                apogee_predictor_data_packet_dict: dict[str, float] = to_builtins(
```
harshil21, line 177: Need a comment specifying that we only care about the last apogee predictor packet

airbrakes/data_handling/logger.py
```diff
@@ -123,55 +126,62 @@ def _convert_unknown_type(obj_type: Any) -> str:
 
     @staticmethod
     def _prepare_log_dict(
```
harshil21, line 128: this function needs to convert the state name to state letter in the context data packet

airbrakes/data_handling/packets/logger_data_packet.py
```diff
@@ -59,5 +59,9 @@ class LoggedDataPacket(TypedDict, total=False):  # total=False means all fields
     vertical_velocity: float | None
     vertical_acceleration: float | None
 
-    # fields not in the data packets
+    # fields in a DebugPacket:
```
harshil21, line 62:
```suggestion
    # fields in a ApogeePredictionPacket:
```

airbrakes/data_handling/packets/logger_data_packet.py
```diff
@@ -59,5 +59,9 @@ class LoggedDataPacket(TypedDict, total=False):  # total=False means all fields
...
     predicted_apogee: float | None
+    uncertainity_threshold_1: str | None
+    uncertainity_threshold_2: str | None
+    fetched_imu_packets: str | None
+    packets_in_imu_queue: str | None
```
harshil21, line 67: all these need to be changed

airbrakes/data_handling/packets/logger_data_packet.py
```diff
@@ -1,16 +1,16 @@
...
+class LoggerDataPacket(TypedDict, total=False):  # total=False means all fields are NotRequired
     """
     Represents a collection of all data that the logger can log in a line. Not every field will be
     filled in every packet.
     """
 
-    state: Required[str]
-    extension: Required[str]
+    state: str
+    extension: str
```
harshil21, line 13: need to add the other ContextPacket fields & add a comment saying that it is of a ContextPacket

harshil21 (changes requested):
core logic looks good.

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -151,6 +173,13 @@ def _curve_fit_function(
         """
         return a * (1 - b * t) ** 4
 
+    def _send_to_main_process(self, curve_coefficients: CurveCoefficients) -> None:
+        """
+        Sends the curve coefficients to the main process.
+        """
+        self.uncertainty_threshold_1.value = curve_coefficients.uncertainties[0]
+        self.uncertainty_threshold_2.value = curve_coefficients.uncertainties[1]
+
```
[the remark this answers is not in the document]
JacksonElia: bro I have no idea i don't remember making this code

harshil21:
One major thing I found - 

https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/blob/be90a4e1a791e63b85fe58e5392755552d87809a/airbrakes/data_handling/logger.py#L173-L180

Since this is asynchronous, we cannot assume that the length of `apogee_predictor_data_packets` is less than the `imu_data_packets` passed to `log()`. Which means there is a super tiny possibility that we don't log all the apogee predictor packets and stop prematurely. 

My recommended fix would be to use another queue to which we add the packets, and then `log()` automatically fetches from it if it's not empty. 

However, I'm first going to test a lot if this would even happen, since we get one ApogeePredictorDataPacket for every 10 ProcessedDataPackets, and we fetch a maximum of 5 EstimatedDataPackets per iteration.

harshil21 (approved):
LGTM. The issue I mentioned does not happen even on 1x speed.

JacksonElia:
@harshil21 that is definitely a good thing to keep in mind, but because of this line in apogee predictor
```py
if len(self._accelerations) - last_run_length >= APOGEE_PREDICTION_MIN_PACKETS:
```

the apogee predictor packets should never out pace the imu packets. Also side note, I think we should consider lowering this to like 5 because even 10 seems a little high (especially considering that we have finally ordered a pi 5).

airbrakes/data_handling/packets/logger_data_packet.py
```diff
@@ -6,17 +6,15 @@
...
-    update_timestamp: int | None
-    state_name: str | None
-    imu_queue_size: int | None
-    apogee_predictor_queue_size: int | None
+    # Field in ContextDataPacket
+    state_letter: str | None
 
     # Fields in ServoDataPacket
-    set_extension: int | None
+    set_extension: str | None
```
JacksonElia, line 17: wait actually shouldn't this be a float or are we logging the enum?

harshil21:
I'm fine with that. Lowering that would make no meaningful difference in convergence time, nor would it affect performance anyway.

We could do it in another PR.

airbrakes/data_handling/packets/logger_data_packet.py
harshil21: we are logging the value, but it will be a string, because if it is a float, our method of truncation will force it to become something like `0.40000000` (8 decimal places truncation)

airbrakes/airbrakes.py
```diff
@@ -87,6 +89,22 @@ def __init__(
...
+    @property
+    def predicted_apogee(self) -> float:
+        """
+        :return: The predicted apogee of the rocket in meters. Returns 0.0 if an predicted
+        apogee is not available.
+        """
+        if self.apogee_predictor_data_packets and (
+            pred_apogee := self.apogee_predictor_data_packets[-1].predicted_apogee
+        ):
+            # Cache the apogee value so we don't have to keep recalculating it
```
JacksonElia, line 104: Why is this necessary? Accessing the list isn't expensive and I don't get what you mean by recalculating. Also I don't like this being a property here, if we make it a property I feel like apogee predictor should own it

airbrakes/data_handling/packets/context_data_packet.py
```diff
@@ -8,11 +10,19 @@ class ContextDataPacket(msgspec.Struct):
     This data packet keeps data owned by the AirbrakesContext as well as metadata about the context.
     """
 
-    update_timestamp: int
+    state_letter: Literal["S", "M", "C", "F", "L"]
```
JacksonElia, line 13: clean

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -97,15 +98,17 @@ def __init__(self):
             self._processed_data_packet_queue: Queue[
                 list[ProcessorDataPacket] | Literal["STOP"]
             ] = Queue(max_size_bytes=BUFFER_SIZE_IN_BYTES)
-            self._apogee_predictor_packet_queue: Queue[ApogeePredictorDataPacket] = Queue()
+            self._apogee_predictor_packet_queue: Queue[ApogeePredictorDataPacket] = Queue(
+                max_size_bytes=BUFFER_SIZE_IN_BYTES
```
JacksonElia, line 102: do we want this to be the same buffer size? also did we put much thought into what the buffer size should be or is it just kinda arbitrary? Not saying thats a bad thing, just curious

airbrakes/data_handling/logger.py
```diff
@@ -172,10 +173,15 @@ def _prepare_log_dict(
...
+            # context packet.
+
+            # There is a small possibility that we won't log all the apogee predictor data packets
+            # if the length of the IMU data packets is less than the length of the apogee predictor
+            # data packets. However, this is unlikely to happen in practice. This particular case
+            # is NOT covered by tests.
             if apogee_predictor_data_packets:
                 apogee_predictor_data_packet_dict: dict[str, float] = to_builtins(
-                    apogee_predictor_data_packets.pop(0)
+                    apogee_predictor_data_packets.pop(0), enc_hook=Logger._convert_unknown_type
```
JacksonElia, line 186: wait what does this do

airbrakes/airbrakes.py
harshil21: This isn't about it being expensive to access, but has two purposes:

1) Provides a clean way for the state machine to query for the apogee
2) `self.apogee_predictor_data_packets` will be an empty list when our loop is only running with a `RawDataPacket`. This means that we'll have an IndexError.
3) Because of 2), we need to store the previous received predicted apogee value. That's what I meant by caching it. 

I'll update the comment to reflect why this is done the way it is.

> property I feel like apogee predictor should own it

fair concern. However this means we also need another init variable in `ApogeePredictor`, and mind you we already have one called `self._apogee`, but that is only referenceable in the other process. So do you want a new name for this or keep in it AirbrakesContext?

airbrakes/data_handling/logger.py
harshil21: converts numpy floats to regular floats

airbrakes/data_handling/packets/logger_data_packet.py
JacksonElia: idk if I like that but I'll address this later if I feel like

airbrakes/data_handling/apogee_predictor.py
harshil21: The default buffer size is a bit too small for our packets specially when we have like over a 100 of them. I remember it completely clogging up the logger queue and it was hard to know why (there was no error either).

The buffer size is big enough to accommodate 1000s of packets

airbrakes/airbrakes.py
JacksonElia: wait i'll just push it I got an idea gimme one min

JacksonElia: lol this is going back on what we said a couple days ago but I'll explain

airbrakes/data_handling/logger.py
JacksonElia: nice

tests/test_apogee_predictor.py
```diff
@@ -168,11 +200,12 @@ def test_prediction_loop_every_x_packets(self, threaded_apogee_predictor):
         # of packets.
         assert 0.0 in unique_apogees
```
wlsanderson, line 201: I'm pretty sure this test should be removed. When we send packets to the apogee predictor, we shouldn't get a zero apogee because the first group of packets in this test causes it to converge, and then that same group of packets will predict apogee with the lookup table.

airbrakes/data_handling/apogee_predictor.py
```diff
@@ -128,17 +142,32 @@ def stop(self) -> None:
...
+        """
+        total_packets = []
+        # get_many doesn't actually get all of the packets, so we need to keep checking until
+        # there are no more packets left
+        with contextlib.suppress(Empty):
+            while True:
+                new_packets = self._apogee_predictor_packet_queue.get_many(block=False)
+                if not new_packets:
+                    break
+                total_packets.extend(new_packets)
```
harshil21, line 169: I don't think we need to do a while loop here, we'll only ever practically get ~<2 packets per iteration

harshil21 (commented):
inital re-review

PR #105, opened 2024-11-25
Add `metadata.json` for launches

### LINKED ISSUE #84: Add a `metadata.json` file in the `launch_data` directory
We need some more context about the launches not only for #76, but also in general. Here are some fields I propose which should be added for every launch:

- Temperature at that day
- Air pressure at that day
- Wind speed + direction at that day 
- Start of log buffer in the file (index)
- Approximate launch site coordinates (I prefer decimal degrees, but fine with whatever)
- Rocket mass
- Type of motor?
- Type of IMU used
- Raw and est packet frequency
- Launch rod angle and direction
- Rocket dimensions?
- etc

This will serve as the ground truth values for the flight and make it easy to load a configuration for a particular flight and then test against those changes and our mock. Thoughts @wlsanderson /@JacksonElia?

harshil21:
Closes #84

### DIFF
--- a/launch_data/metadata.json
+++ b/launch_data/metadata.json
@@ -21,7 +21,7 @@
             "log_buffer_index": 5000,
             "apogee_meters": 457.68
         },
-        "flight_description": "Unknown."
+        "flight_description": "The servo died right before the launch, so this was just a data collection flight. It was with the first iteration of the Airbrakes code, so it doesn't have all of the fields we normally log."
     },
     "interest_launch.csv": {
         "date": "2024-08-28",
--- a/launch_data/metadata.json
+++ b/launch_data/metadata.json
@@ -1,5 +1,6 @@
 {
     "purple_launch.csv": {
+        "date": "2023-12-16",
         "launch_site": {
             "air_temperature_celsius": 16.11,  
             "air_pressure_mb": 1031.49,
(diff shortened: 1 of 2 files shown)


launch_data/metadata.json
```diff
@@ -0,0 +1,99 @@
+{
+    "purple_launch.csv": {
+        "date": "2023-12-16",
```
harshil21, line 3: need to add the time of launch as well

JacksonElia (approved):
This looks good but what is the point of it? Will it be used in the new UI you're working on?

harshil21:
Yes, and anyone who has no idea about the launch will get an idea of what happened.

launch_data/metadata.json
```diff
@@ -0,0 +1,99 @@
...
+        "imu_details": {
+            "imu_model": "Parker-LORD 3DM-CX5-15",
+            "raw_packet_frequency": 1000.0,
+            "est_packet_frequency": 500.0
+        },
+        "flight_data": {
+            "log_buffer_index": 5000,
+            "apogee_meters": 457.68
+        },
+        "flight_description": "Unknown."
```
harshil21, line 24: @JacksonElia since you were there last year, some context would be nice

PR #107, opened 2024-11-25
Having rotated accelerations use quaternions instead of gyroscope

wlsanderson:
Because we have good datasets now, we can have the rotated accelerations be calculated with quaternions directly and not from gyro. We will have gyro as a backup, and if any of the uncertainties of the quaternions go above some threshold (in constants file), the data processor will calculate with gyroscope as a fallback. This change should improve the speed of this function by a lot, as it can vectorize a large portion of it.

### DIFF
--- a/airbrakes/data_handling/data_processor.py
+++ b/airbrakes/data_handling/data_processor.py
@@ -8,7 +8,7 @@ from scipy.spatial.transform import Rotation as R
 
 from airbrakes.data_handling.imu_data_packet import EstimatedDataPacket
 from airbrakes.data_handling.processed_data_packet import ProcessedDataPacket
-from constants import ACCELERATION_NOISE_THRESHOLD, GRAVITY
+from constants import ACCELERATION_NOISE_THRESHOLD, GRAVITY, QUATERNION_UNCERTAINTY_THRESHOLD
 from utils import deadband
 
 
@@ -202,52 +202,98 @@ class IMUDataProcessor:
     def _calculate_rotated_accelerations(self) -> npt.NDArray[np.float64]:
         """
         Calculates the rotated acceleration vector. Converts gyroscope data into a delta
-        quaternion, and adds onto the last quaternion. Will most likely be replaced by IMU
-        quaternion data in the future, this is a work-around due to bad datasets.
+        quaternion, and adds onto the last quaternion. Only used when the quaternions are
+        deemed too uncertain to rely on.
 
-        :return: numpy list of rotated acceleration vector [x,y,z]
+        :return: numpy list of the vertical rotated accelerations
         """
-
         # We pre-allocate the space for our accelerations first
         rotated_accelerations = np.zeros(len(self._data_packets))
 
-        current_orientation = self._current_orientation_quaternions
-        # Iterates through the data points and time differences between the data points
-        for idx, (data_packet, dt) in enumerate(
-            zip(self._data_packets, self._time_differences, strict=True)
-        ):
-            # Accelerations are in m/s^2
-            x_accel = data_packet.estCompensatedAccelX
-            y_accel = data_packet.estCompensatedAccelY
-            z_accel = data_packet.estCompensatedAccelZ
-            # Angular rates are in rads/s
-            gyro_x = data_packet.estAngularRateX
-            gyro_y = data_packet.estAngularRateY
-            gyro_z = data_packet.estAngularRateZ
+        # get the compensated accelerations from each data packet
+        compensated_accelerations = np.array(
+            [
+                [
+                    packet.estCompensatedAccelX,
+                    packet.estCompensatedAccelY,
+                    packet.estCompensatedAccelZ,
+                ]
+                for packet in self._data_packets
+            ]
+        )
+        # check if the quaternions are too uncertain or not. If they are not, we calculate
+        # the rotated accelerations directly with the quaternion data, if they are uncertain,
+        # we calculate the quaternions from the gyroscope data, and then find the rotations.
+        last_packet = self._data_packets[-1]
+        quaternion_uncertainties = np.array(
+            [
+                last_packet.estAttitudeUncertQuaternionW,
+                last_packet.estAttitudeUncertQuaternionX,
+                last_packet.estAttitudeUncertQuaternionY,
+                last_packet.estAttitudeUncertQuaternionZ,
+            ]
+        )
+        if np.any(quaternion_uncertainties >= QUATERNION_UNCERTAINTY_THRESHOLD):
+            # quaternions are too uncertain to use, so we calculate them from gyro
 
-            # Check for missing data points
-            if any(val is None for val in [x_accel, y_accel, z_accel, gyro_x, gyro_y, gyro_z]):
-                return rotated_accelerations
+            current_orientation = self._current_orientation_quaternions
+            # Iterates through the data points and time differences between the data points
+            for idx, (data_packet, dt) in enumerate(
+                zip(self._data_packets, self._time_differences, strict=True)
+            ):
+                # Accelerations are in m/s^2
+                x_accel = compensated_accelerations[idx][0]
+                y_accel = compensated_accelerations[idx][1]
+                z_accel = compensated_accelerations[idx][2]
+                # Angular rates are in rads/s
+                gyro_x = data_packet.estAngularRateX
+                gyro_y = data_packet.estAngularRateY
+                gyro_z = data_packet.estAngularRateZ
 
-            # scipy docs for more info: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html
-            # Calculate the delta quaternion from the angular rates
-            delta_rotation = R.from_rotvec([gyro_x * dt, gyro_y * dt, gyro_z * dt])
+                # Check for missing data points
+                if any(val is None for val in [x_accel, y_accel, z_accel, gyro_x, gyro_y, gyro_z]):
+                    return rotated_accelerations
 
-            # Update the current orientation by applying the delta rotation
-            current_orientation = current_orientation * delta_rotation
+                # scipy docs for more info: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html
+                # Calculate the delta quaternion from the angular rates
+                delta_rotation = R.from_rotvec([gyro_x * dt, gyro_y * dt, gyro_z * dt])
 
-            # Rotate the acceleration vector using the updated orientation
-            rotated_accel = current_orientation.apply([x_accel, y_accel, z_accel])
-            # Vertical acceleration will always be the 3rd element of the rotated vector,
-            # regardless of orientation. For simplicity, we multiply by -1 so that acceleration
-            # during motor burn is positive, and acceleration due to drag force during coast phase
-            # is negative.
-            rotated_accelerations[idx] = -rotated_accel[2]
+                # Update the current orientation by applying the delta rotation
+                current_orientation = current_orientation * delta_rotation
+
+                # Rotate the acceleration vector using the updated orientation
+                rotated_accel = current_orientation.apply([x_accel, y_accel, z_accel])
+                # Vertical acceleration will always be the 3rd element of the rotated vector,
+                # regardless of orientation. For simplicity, we multiply by -1 so that acceleration
+                # during motor burn is positive, and acceleration due to drag force during coast
+                # phase is negative.
+                rotated_accelerations[idx] = -rotated_accel[2]
+
+            # Update the class attribute with the latest quaternion orientation
+            self._current_orientation_quaternions = current_orientation
+
+            return rotated_accelerations
+
+        # the quaternions are not uncertain, so we can directly use the quaternions given from
+        # the IMU.
+        current_orientation = np.array(
+            [
+                last_packet.estOrientQuaternionW,
+                last_packet.estOrientQuaternionX,
+                last_packet.estOrientQuaternionY,
+                last_packet.estOrientQuaternionZ,
+            ]
+        )
+        # rotate each compensated acceleration by the most recent quaternion orientation. This
+        # introduces some slight error, because the orientation for the last packet is not going
+        # to be the orientation for each packet, but it should be a negligible difference.
+        current_orientation = R.from_quat(current_orientation, scalar_first=True)
+        rotated_accels = current_orientation.apply(compensated_accelerations)
 
         # Update the class attribute with the latest quaternion orientation
         self._current_orientation_quaternions = current_orientation
-
-        return rotated_accelerations
+        # we only want to return the vertical component of each acceleration
+        return np.array([accel[2] for accel in rotated_accels])
 
     def _calculate_vertical_velocity(self) -> npt.NDArray[np.float64]:
         """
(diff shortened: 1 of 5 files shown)


JacksonElia:
Do you think that the quaternions will be more accurate than using the gyroscope? I'm not too well versed in this but don't the quaternions have the potential to accumulate error? I guess because its just accel though it doesn't really matter.

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-        current_orientation = self._current_orientation_quaternions
-        # Iterates through the data points and time differences between the data points
-        for idx, (data_packet, dt) in enumerate(
-            zip(self._data_packets, self._time_differences, strict=True)
+        # checks for missing data
+        if any(
+            packet.estCompensatedAccelX is None
+            or packet.estCompensatedAccelY is None
+            or packet.estCompensatedAccelZ is None
+            for packet in self._data_packets
```
harshil21, line 218:
```suggestion
```
fair to say if any of these are missing we are cooked. Better to just not include it it then.

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-            current_orientation = current_orientation * delta_rotation
-
-            # Rotate the acceleration vector using the updated orientation
-            rotated_accel = current_orientation.apply([x_accel, y_accel, z_accel])
-            # Vertical acceleration will always be the 3rd element of the rotated vector,
-            # regardless of orientation. For simplicity, we multiply by -1 so that acceleration
-            # during motor burn is positive, and acceleration due to drag force during coast phase
-            # is negative.
-            rotated_accelerations[idx] = -rotated_accel[2]
+            return rotated_accelerations
```
harshil21, line 220:
```suggestion
```
see above. github doesn't allow me to give suggestions on deleted lines so

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+                last_packet.estAttitudeUncertQuaternionW,
+                last_packet.estAttitudeUncertQuaternionX,
+                last_packet.estAttitudeUncertQuaternionY,
+                last_packet.estAttitudeUncertQuaternionZ,
+            ]
+        )
+
+        if (
+            np.any(quaternion_uncertainties >= QUATERNION_UNCERTAINTY_THRESHOLD)
+            or self._data_packets[-1].estOrientQuaternionW is None
```
harshil21, line 249: why is only the W quaternion checked for None? To account for missing quaternions in interest launch data? Then imo we should modify that dataset also, to add a random value, maybe exactly 0.0. We should also add `invalid_fields` for that dataset.

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+        ):
+            # quaternions are too uncertain to use, so we calculate them from gyro
```
harshil21, line 251: calculating from gyro data should be in a different method

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+            [
+                last_packet.estOrientQuaternionW,
+                last_packet.estOrientQuaternionX,
+                last_packet.estOrientQuaternionY,
+                last_packet.estOrientQuaternionZ,
+            ]
+        )
+        # rotate each compensated acceleration by the most recent quaternion orientation. This
+        # introduces some slight error, because the orientation for the last packet is not going
+        # to be the orientation for each packet, but it should be a negligible difference.
```
harshil21, line 302: after #99 the error would be even lower, but I would be interested in what the overall difference in predicted apogee is, when running the sim with `-f`.

tests/test_data_processor.py
```diff
@@ -349,6 +401,10 @@ def test_first_update_no_data_packets(self, data_processor):
                         estOrientQuaternionX=0.2,
                         estOrientQuaternionY=0.3,
                         estOrientQuaternionZ=0.4,
+                        estAttitudeUncertQuaternionW=0.01,
+                        estAttitudeUncertQuaternionX=0.01,
+                        estAttitudeUncertQuaternionY=0.01,
+                        estAttitudeUncertQuaternionZ=0.01,
```
harshil21, line 407: lol we really need a helper method to generate data packets like we have in test_logger, this must've been atrocious to edit

tests/test_data_processor.py
```diff
@@ -499,11 +555,58 @@ def test_max_altitude(self, data_processor):
...
+        data_packets = load_data_packets(csv_path, n_packets)
+        d = IMUDataProcessor()
+        for i, _ in enumerate(data_packets):
+            data_packets[i].estAttitudeUncertQuaternionW = 1
+        print(data_packets)
+        d.update(data_packets)
+        rotations = d._rotated_accelerations
+        assert len(rotations) == n_packets
+
+        assert rotations[-1] == pytest.approx(expected_value)
```
harshil21, line 612: Interesting that the expected_value is the same for both methods. I would've expected a small difference at least.

tests/test_data_processor.py
```diff
@@ -499,11 +555,58 @@ def test_max_altitude(self, data_processor):
...
     )
-    def test_calculate_rotations(self, csv_path, expected_value, n_packets):
+    def test_calculate_rotations_quaternions(self, csv_path, expected_value, n_packets):
         data_packets = load_data_packets(csv_path, n_packets)
         d = IMUDataProcessor()
         d.update(data_packets)
         rotations = d._rotated_accelerations
         assert len(rotations) == n_packets
 
         assert rotations[-1] == pytest.approx(expected_value)
```
harshil21, line 565:
```suggestion
        assert rotations[-1] == pytest.approx(expected_value)
        assert rotations[0] == rotations[-1]
```
this checks that all the rotations set were the same for that list for the quaternion only case.

harshil21 (changes requested):
I think we discussed this briefly, but wouldn't it be safer to just check for `invalid_fields` rather than uncertainties?

airbrakes/data_handling/data_processor.py
```diff
@@ -202,52 +202,111 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
+        current_orientation = R.from_quat(current_orientation, scalar_first=True)
+        rotated_accels = current_orientation.apply(compensated_accelerations)
 
         # Update the class attribute with the latest quaternion orientation
         self._current_orientation_quaternions = current_orientation
-
-        return rotated_accelerations
+        # we only want to return the negative vertical component of each acceleration
+        return np.array([-accel[2] for accel in rotated_accels])
```
harshil21, line 309:
```suggestion
        return -rotated_accels[:, -1]
```
This is **95.1%** faster since it's vectorized!!

```python
import timeit

setup = """
import numpy as np

n = np.array([[1.213, 1.213 + i, 1.213 +2*i] for i in range(100)])
"""

test_1 = """
b = -n[:, -1]
"""

test_2 = """
b = np.array([-j[-1] for j in n])
"""

time_1 = timeit.timeit(test_1, setup, number=1000000)
time_2 = timeit.timeit(test_2, setup, number=1000000)

print(time_1)
print(time_2)
```
```bash
└─[$] <git:(adjust-queue-fetching*)> python scripts/perf.py
0.6513239109990536
24.606471167000564
```

wlsanderson:
> I think we discussed this briefly, but wouldn't it be safer to just check for `invalid_fields` rather than uncertainties?

No, the issue with invalid fields is that it is not flexible enough. From the previous datasets, it seemed that quaternions became invalid when an uncertainty component was around 0.1, but usually stayed invalid for multiple seconds, even when the uncertainty dropped back significantly below 0.1. Being able to tweak this value ourselves by just using the uncertainty is much more versatile. 0.1 is also incredibly sensitive, and I think increasing this to 0.5 is still well within a safe margin. We have only seen issues with quaternion uncertainty when it had that bug and hit a billion.

Even in freefall when some of the components of quaternion uncertainty reached over 2.0, the data is still good enough for us to get accurate (+/- 20 m/s) velocity tracking once it hits the ground.

airbrakes/data_handling/data_processor.py
wlsanderson: Yeah I was kinda debating this idea with what to check when running this function. It's not like we have ever encountered a situation where a packet doesn't have acceleration or gyroscope, so idk if it is even worth checking

airbrakes/data_handling/data_processor.py
wlsanderson: yeah I'll go back and flesh it out a bit better. You think I should just check for missing quaternions and missing quaternion uncertainties only? Don't really know if checking for missing accel/gyro is worth it

airbrakes/data_handling/data_processor.py
wlsanderson: Gotcha, should this logic of determining which calculation to use be in the data_processor update method? or should the update method call quaternion rotation method, and that will call the gyro rotation method if uncertainty is too high?

airbrakes/data_handling/data_processor.py
wlsanderson: man I was sure there was a way to vectorize this but couldn't figure it out lol

wlsanderson:
> Do you think that the quaternions will be more accurate than using the gyroscope? I'm not too well versed in this but don't the quaternions have the potential to accumulate error? I guess because its just accel though it doesn't really matter.

I'm not exactly sure how the IMU does this, but the quaternions are absolute measurements, not relative. Basically, the quaternions always give the rotation with respect to some fixed axis, regardless of what the initial orientation of the IMU was when it was powered on. Because of this, I'd imagine quaternions are less susceptible to gyro drift. Gyroscope is relative, and so we just have to use dead reckoning to estimate our orientation.

airbrakes/data_handling/data_processor.py
harshil21: I don't think we should check any of that, they are always guaranteed to be there. Let's just modify the older datasets

airbrakes/data_handling/data_processor.py
harshil21: Yeah we have enough launches with this IMU to know that's not going to happen

airbrakes/data_handling/data_processor.py
harshil21: the latter. the update() method is a higher level method and doesn't need to be concerned with the specifics

airbrakes/data_handling/data_processor.py
harshil21: https://numpy.org/devdocs/user/basics.rec.html#structured-arrays

oh my god look what I found

wlsanderson: if we did this, would msgspec not be needed?

harshil21: it's a bit early to say that, but I'm looking at alternative ways of representing data, like pandas Dataframes (which are also labeled). It's just that it hurts the readability of having data packets to represent data v/s Dataframes, but the latter is way more suitable for us since we keep working on the columns (e.g. data processor and apogee predictor) rather than the row (only used in logging)

harshil21: Resolved in #112

wlsanderson:
Still a work in progress, but it is possible that our gyroscope calculations were wrong the entire time. I don't really know how. I added quaternion uncertainty to purple nurple, and added the missing quaternions to interest launch. When checking interest launch, it turns out that during the beginning of coast phase with gyroscope calculations that we are currently doing on the main branch, we are at a small pitch angle of around 5-10 degrees. The original quaternions (the ones that actually logged in the original flight) are at a pitch angle of around 20 - 30 degrees at the same point in flight.

I'm not actually sure which one is correct. Its possible that the gyroscope data does not line up with the orientation of the rocket, or it is possible that in the interest launch data when quaternion data became "valid" again, it had a large amount of error. I will do some checking with genesis flights to see how those compare with pure quaternion  vs pure gyro, but it might be different enough between full-scale and sub-scale that genesis won't be a good comparison. If this is true though, this might be the reason why #76 apogee prediction convergence time doesn't work for only full scale, but not sub scale. Thoughts? @JacksonElia @harshil21

wlsanderson:
After thinking about it, i feel like I might know why this is. I'm not sure which one, but I think either integrating gyro, or IMU quaternions, actually give the orientation of the rocket (which way it is facing), and one of them gives the direction it is travelling (direction of the acceleration vector). The angle between these two might be the angle of attack of the rocket

harshil21:
Interesting observations. This is why I wanted some kind of a comparison between the two approaches. Do you think that integrating gyro data (which is prone to drift) is causing a large error between the two? What does the plot of the pitch from T+0 to some time in coast phase look like for gyro data only v/s quaternion data only? (I guess only for genesis since we actually logged that).

I would bias towards quaternions being more accurate since that's what they actually represent the absolute orientation.

wlsanderson:
I'm going to test all that stuff today.

> I would bias towards quaternions being more accurate since that's what they actually represent the absolute orientation.

That's what I originally thought too, but I dm'd Sailor last night and he said that the reason it seems absolute is most likely because it uses something like the gravity vector to initialize the quaternion, and then iterate on that with gyroscope. So it's "absolute" for the first packet when the IMU boots up, and then uses that as a reference axis for relative measurements afterwards.

I'm going to test more though to see what is most reasonable. I definitely want to consider getting a standalone magnetometer though, for the IMU to use as an external heading system.

harshil21:
If you/we have time, we can go over the mscl library code, then we'd actually know what they are doing and e.g. when quaternions become invalid instead of just guessing

wlsanderson:
sure, i'm available all today in the lab to work on it

wlsanderson:
Closing this PR because after reviewing interest launch, and both genesis flight data, I believe that determining orientation from gyroscope is still the most robust and consistent method. The branch will remain available because we might return to it in the future, if a magnetometer proves to fix this issue.

I graphed the pitch versus time for interest launch, taking note of the regions (gray area) that quaternions became invalid. I graphed the pitch in two ways: integrating the gyroscope how we currently do it (blue), and using the IMU's quaternions to get orientation instead (red).
pic1

It's clear that there are two notable "jumps" in the quaternion orientation, while the gyroscope orientation stays steady, as expected. While it's hard to draw many conclusions from the larger jump because it happened during an invalid section where we did not log data, the first jump happens around a tenth of a second before quaternions are marked invalid, so this was a quaternion with an unnatural jump that the IMU returned. The packets themselves are shown below:
image

Both genesis launches did not exhibit any jumps similar to these during motor burn or coast phase, but they existed in freefall state. Also, take note of the precision between the two measurements up until the first jump in data.
pic2
pic3

Despite not having any quaternion jumps during coast phase, I believe that this was either just luck, or a result of being smaller (slower, shorter) flights. We also cannot use quaternion uncertainty to validate quaternions, because these jumps occurred during relatively normal uncertainty levels, and they never corrected themselves either, the error only accumulated.

I think it is possible that a magnetometer could completely stop these jumps, or allow for some internal correcting in the IMU readings. Although, until we can get launch data with a magnetometer connected, I think quaternions are too volatile to use and we should stick with integrating gyroscope. There is an argument to be made for checking to see if the quaternions have jumped and then using gyroscope for the rest of the flight, but it seems like more effort than what it is worth, especially because performance has increased a lot and that would be the main benefit of not integrating gyroscope.

Issue #108, opened 2024-11-27
Code needs to be cleaned up more


JacksonElia:
Honestly our code is pretty good at this point, but it could always be better. We need to:
- [x] Clean up `constants.py`
- [ ] Add comments to the tests (oh boy)
- [ ] Do another go at changing var names and function names to be clearer and convey more info
- [ ] Flesh out `scripts/` a little more, maybe add a nicer UI for them or something
- [ ] Expand the readme, maybe add a little more documentation/guidance where to start for new members

harshil21:
Duplicate of #38

wlsanderson:
> Duplicate of #38

so real

harshil21:
> Flesh out scripts/ a little more, maybe add a nicer UI for them or something

This can probably be done along with/after #106.

PR #106, opened 2024-11-25
Massive display overhaul

harshil21:
Biggest PR in airbrakes history. Provides a sleek and eye-catching TUI. It is recommended to use a modern terminal like kitty for the best experience (performance + graphics wise). 

There are still many more features, changes, and tests I would like to add, but for a first iteration, we have made it.

To try it out, simply run `uv run mock`. 

Closes #54.

Some non-display related changes, which was required for reliability reasons with the display:

- Forces multiprocessing start method to be "spawn" for all processes. "forkserver" does not work with `textual-image` unfortunately.
- `BaseIMU` now has a new shared variable called `request_to_start`, which is different from `is_running`, which is set only when the process is actually running. This distinction helps achieve cleaner starts and stops. 
- A new `ExtendedDataProcessor` class.
- It is currently not possible to run a `sim`. 

----
_Original post:_

To do:

- [x] Add T- launch time
- [x] Read all of the textual docs to understand how best to integrate and make the display
- [x] Barebones: Get display working with mock imu
- [x] Start adding graphs
- [x] Add launch file selector with some options, with ~picture of the rocket~ (the picture of the rocket is too low res for the terminal) along with relevant metadata
- [x] Add downrange distance calculation in the (x,y) plane in meters, it should show live trajectory, and estimated trajectory.
- [x] Add 3D model showing orientation of the rocket (not possible with textual, probably not worth using another library? Maybe a 2D model would have to suffice...)
- [ ] Detect drogue and main chute deployment
- [x] Add debug panel for information (apparently only possible with a 2nd terminal)
- [x] Integrate simulation speed into MockIMU
- [x] Create main and ~sim~ launch screen
- [x] Benchmark mode in mock - No display, just the main loop and show flight stats after benchmark.
- [x] Sound alarm if IMU has issues
- [ ] Document things
- [x] Make everything pretty
- [ ] Tests with textual

Estimated lines of code to add: 1000+ python, probably at least a couple hundred for the CSS.

PR #109, opened 2024-11-28
More Cleanup

JacksonElia:
This is mainly to address #108 

The more I do this, the more necessary I realize it is. I changed like 5 comments/names that were just completely incorrect.

### DIFF
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -56,7 +56,7 @@ class AirbrakesContext:
         :param servo: The servo object that controls the extension of the airbrakes. This can be a
         real servo or a mock servo.
         :param imu: The IMU object that reads data from the rocket's IMU. This can be a real IMU or
-            a mock IMU.
+        a mock IMU.
         :param logger: The logger object that logs data to a CSV file.
         :param data_processor: The data processor object that processes IMU data on a higher level.
         :param apogee_predictor: The apogee predictor object that predicts the apogee of the rocket.
--- a/airbrakes/data_handling/logger.py
+++ b/airbrakes/data_handling/logger.py
@@ -94,16 +94,54 @@ class Logger:
         return f"{obj_type:.8f}"
 
     @staticmethod
-    def _truncate_floats(data: LoggedDataPacket) -> dict[str, str | object]:
+    def _prepare_log_dict(
+        state: str,
+        extension: str,
+        imu_data_packets: deque[IMUDataPacket],
+        processed_data_packets: deque[ProcessedDataPacket],
+        predicted_apogee: float,
+    ) -> deque[LoggedDataPacket]:
         """
-        Truncates the decimal place of the floats in the dictionary to 8 decimal places.
-        :param data: The dictionary to truncate.
-        :return: The truncated dictionary.
+        Creates a data packet dictionary representing a row of data to be logged.
+        :param state: The current state of the airbrakes.
+        :param extension: The current extension of the airbrakes.
+        :param imu_data_packets: The IMU data packets to log.
+        :param processed_data_packets: The processed data packets to log.
+        :param: predicted_apogee: The predicted apogee of the rocket. Only logged with estimated
+        data packets.
+        :return: A deque of LoggedDataPacket objects.
         """
-        return {
-            key: f"{value:.8f}" if isinstance(value, float) else value
-            for key, value in data.items()
-        }
+        logged_data_packets: deque[LoggedDataPacket] = deque()
+
+        # Convert the imu data packets to a dictionary:
+        for imu_data_packet in imu_data_packets:
+            # Let's first add the state, extension field:
+            logged_fields: LoggedDataPacket = {
+                "state": state,
+                "extension": extension,
+            }
+            # Convert the imu data packet to a dictionary
+            # Using to_builtins() is much faster than asdict() for some reason
+            imu_data_packet_dict: dict[str, int | float | list[str]] = to_builtins(imu_data_packet)
+            logged_fields.update(imu_data_packet_dict)
+
+            # Get the processed data packet fields:
+            if isinstance(imu_data_packet, EstimatedDataPacket):
+                # Convert the processed data packet to a dictionary. Unknown types such as numpy
+                # float64 are converted to strings with 8 decimal places (that's enc_hook)
+                processed_data_packet_dict: dict[str, float] = to_builtins(
+                    processed_data_packets.popleft(), enc_hook=Logger._convert_unknown_type
+                )
+                # Let's drop the "time_since_last_data_packet" field:
+                processed_data_packet_dict.pop("time_since_last_data_packet", None)
+                logged_fields.update(processed_data_packet_dict)
+
+                # Add the predicted apogee field:
+                logged_fields["predicted_apogee"] = predicted_apogee
+
+            logged_data_packets.append(logged_fields)
+
+        return logged_data_packets
 
     def start(self) -> None:
         """
@@ -136,13 +174,13 @@ class Logger:
         :param imu_data_packets: The IMU data packets to log.
         :param processed_data_packets: The processed data packets to log.
         :param predicted_apogee: The predicted apogee of the rocket. Note: Since this is fetched
-            from the apogee predictor, which runs in a separate process, it may not be the most
-            recent value, nor would it correspond to the respective data packet being logged.
+        from the apogee predictor, which runs in a separate process, it may not be the most recent
+        value, nor would it correspond to the respective data packet being logged.
         """
         # We only want the first letter of the state to save space
         state_letter = state[0]
         # We are populating a dictionary with the fields of the logged data packet
-        logged_data_packets: deque[LoggedDataPacket] = self._prepare_log_dict(
+        logged_data_packets: deque[LoggedDataPacket] = Logger._prepare_log_dict(
             state_letter,
             str(extension),
             imu_data_packets,
@@ -180,55 +218,18 @@ class Logger:
             self._log_queue.put(buffered_message)
         self._log_buffer.clear()
 
-    def _prepare_log_dict(
-        self,
-        state: str,
-        extension: str,
-        imu_data_packets: deque[IMUDataPacket],
-        processed_data_packets: deque[ProcessedDataPacket],
-        predicted_apogee: float,
-    ) -> deque[LoggedDataPacket]:
+    # ------------------------ ALL METHODS BELOW RUN IN A SEPARATE PROCESS -------------------------
+    @staticmethod
+    def _truncate_floats(data: LoggedDataPacket) -> dict[str, str | object]:
         """
-        Creates a data packet dictionary representing a row of data to be logged.
-        :param state: The current state of the airbrakes.
-        :param extension: The current extension of the airbrakes.
-        :param imu_data_packets: The IMU data packets to log.
-        :param processed_data_packets: The processed data packets to log.
-        :param: predicted_apogee: The predicted apogee of the rocket. Only logged with estimated
-            data packets.
-        :return: A deque of LoggedDataPacket objects.
+        Truncates the decimal place of the floats in the dictionary to 8 decimal places.
+        :param data: The dictionary to truncate.
+        :return: The truncated dictionary.
         """
-        logged_data_packets: deque[LoggedDataPacket] = deque()
-
-        # Convert the imu data packets to a dictionary:
-        for imu_data_packet in imu_data_packets:
-            # Let's first add the state, extension field:
-            logged_fields: LoggedDataPacket = {
-                "state": state,
-                "extension": extension,
-            }
-            # Convert the imu data packet to a dictionary
-            # Using to_builtins() is much faster than asdict() for some reason
-            imu_data_packet_dict: dict[str, int | float | list[str]] = to_builtins(imu_data_packet)
-            logged_fields.update(imu_data_packet_dict)
-
-            # Get the processed data packet fields:
-            if isinstance(imu_data_packet, EstimatedDataPacket):
-                # Convert the processed data packet to a dictionary. Unknown types such as numpy
-                # float64 are converted to strings with 8 decimal places (that's enc_hook)
-                processed_data_packet_dict: dict[str, float] = to_builtins(
-                    processed_data_packets.popleft(), enc_hook=Logger._convert_unknown_type
-                )
-                # Let's drop the "time_since_last_data_packet" field:
-                processed_data_packet_dict.pop("time_since_last_data_packet", None)
-                logged_fields.update(processed_data_packet_dict)
-
-                # Add the predicted apogee field:
-                logged_fields["predicted_apogee"] = predicted_apogee
-
-            logged_data_packets.append(logged_fields)
-
-        return logged_data_packets
+        return {
+            key: f"{value:.8f}" if isinstance(value, float) else value
+            for key, value in data.items()
+        }
 
     def _logging_loop(self) -> None:
         """
--- a/airbrakes/data_handling/logger.py
+++ b/airbrakes/data_handling/logger.py
@@ -24,8 +24,6 @@ class Logger:
 
     It uses Python's csv module to append the airbrakes' current state, extension, and IMU data to
     our logs in real time.
-
-    :param log_dir: The directory where the log files will be.
     """
 
     __slots__ = ("_log_buffer", "_log_counter", "_log_process", "_log_queue", "log_path")
@@ -94,6 +92,18 @@ class Logger:
         :return: The truncated object.
         """
         return f"{obj_type:.8f}"
+
+    @staticmethod
+    def _truncate_floats(data: LoggedDataPacket) -> dict[str, str | object]:
+        """
+        Truncates the decimal place of the floats in the dictionary to 8 decimal places.
+        :param data: The dictionary to truncate.
+        :return: The truncated dictionary.
+        """
+        return {
+            key: f"{value:.8f}" if isinstance(value, float) else value
+            for key, value in data.items()
+        }
 
     def start(self) -> None:
         """
@@ -207,7 +217,7 @@ class Logger:
                 # Convert the processed data packet to a dictionary. Unknown types such as numpy
                 # float64 are converted to strings with 8 decimal places (that's enc_hook)
                 processed_data_packet_dict: dict[str, float] = to_builtins(
-                    processed_data_packets.popleft(), enc_hook=self._convert_unknown_type
+                    processed_data_packets.popleft(), enc_hook=Logger._convert_unknown_type
                 )
                 # Let's drop the "time_since_last_data_packet" field:
                 processed_data_packet_dict.pop("time_since_last_data_packet", None)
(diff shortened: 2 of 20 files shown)


constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
 
-# The "IDs" of the data packets that the IMU sends
+# The "IDs" of the data packets that the IMU sends.
 ESTIMATED_DESCRIPTOR_SET = 130
+"""The ID of the estimated data packet that the IMU sends"""
 RAW_DESCRIPTOR_SET = 128
+"""The ID of the raw data packet that the IMU sends"""
 
-# The maximum size of the data queue for the packets, so we don't run into memory issues
 MAX_QUEUE_SIZE = 100000
```
harshil21, line 63:
```suggestion
MAX_QUEUE_SIZE = 100_000
```

constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
-# The maximum amount of time in seconds the IMU process will wait for a packet before timing out:
-PROCESS_TIMEOUT = 3
-
-# -------------------------------------------------------
-# Data Processing Configuration
-# -------------------------------------------------------
-
+IMU_TIMEOUT_SECONDS = 3.0
+"""The maximum amount of time in seconds the IMU process to do something (e.g. read a packet) before
+it is considered to have timed out. This is used to prevent the program from hanging if the IMU
```
harshil21, line 72:
```suggestion
it is considered to have timed out. This is used to prevent the program from deadlocking if the IMU
```

constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
-# The path of the folder to hold the log files in
 LOGS_PATH = Path("logs")
+"""The path of the folder to hold the log files in"""
 TEST_LOGS_PATH = Path("test_logs")
+"""The path of the folder to hold the test log files in"""
 
-# The signal to stop the logging process, this will be put in the queue to stop the process
-# see stop() and _logging_loop() for more details.
 STOP_SIGNAL = "STOP"
+"""The signal to stop the logging process, this will be put in the queue to stop the process"""
```
harshil21, line 85:
```suggestion
"""The signal to stop the logging and the apogee prediction process, this will be put in the queue to stop the process"""
```

constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
-LANDED_SPEED = 5.0  # m/s
-
+"""Because motors can behave unpredictably near the end of their burn, we will only say that the
+motor has stopped burning if the current velocity <= Max velocity * (1 - MAX_VELOCITY_THRESHOLD)"""
+
+# ----------------- Freefall to Landing -----------------
+MAX_FREE_FALL_SECONDS = 300.0
+"""The maximum amount of time in seconds that the rocket can be in freefall before we consider it to
+have landed. This is to prevent the program from running indefinitely if our code never detects the 
+landing of the rocket. This value accounts for the worst case scenario of the main parachute
```
harshil21, line 127: definitely saved us last launch

constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
-# Free fall to Landing:
-MAX_FREE_FALL_LENGTH = 180.0  # seconds
-
-# Consider the rocket to have landed if it is within 15 meters of the launch site height
-# and the speed is low.
-GROUND_ALTITUDE = 10.0  # meters
-LANDED_SPEED = 5.0  # m/s
-
+"""Because motors can behave unpredictably near the end of their burn, we will only say that the
+motor has stopped burning if the current velocity <= Max velocity * (1 - MAX_VELOCITY_THRESHOLD)"""
```
harshil21, line 121: this 1 minus thing is weird. Why not just change this to 0.96?

constants.py
```diff
@@ -45,99 +49,117 @@ class DisplayEndingType(StrEnum):
...
+"""The initial guess for the coefficients for curve fit of the acceleration curve."""
+
+APOGEE_PREDICTION_MIN_PACKETS = 10
+"""The minimum number of data packets required to predict the apogee."""
 
-# The uncertainty from the curve fit, below which we will say that our apogee has converged:
-UNCERTAINTY_THRESHOLD = [0.0359, 0.00075]  # [0.0259, 0.00065]
-# UNCERTAINTY_THRESHOLD = [0.1, 0.75]  # [0.0259, 0.00065]
+UNCERTAINTY_THRESHOLD = [0.0359, 0.00075]  # For near quick convergence times, use: [0.1, 0.75]
+"""he uncertainty from the curve fit, below which we will say that our apogee has converged"""
```
harshil21, line 165:
```suggestion
"""The uncertainty from the curve fit, below which we will say that our apogee has converged"""
```

constants.py
JacksonElia: I like saying that it is within 4 percent

airbrakes/mock/mock_imu.py
```diff
@@ -157,3 +157,6 @@ def _fetch_data_loop(
         # function in a context manager to suppress the KeyboardInterrupt
         with contextlib.suppress(KeyboardInterrupt):
             self._read_file(log_file_path, real_time_simulation, start_after_log_buffer)
+
+        # For the mock, once we're done reading the file, we say it is no longer running
+        self._running.value = False
```
harshil21, line 162: nice, why didn't I think of that

constants.py
harshil21: Ok, how about `MAX_VELOCITY_THRESHOLD = 1 - 0.04`? You see it's just an unnecessary computation every time

JacksonElia: it being an unnecessary computation is bull, that shit probably is so negligible but you're right, makes it more complex/hard to read

airbrakes/mock/mock_imu.py
JacksonElia: lol prolly just a 3am coding thing

airbrakes/hardware/imu.py
```diff
@@ -123,69 +184,27 @@ def _query_imu_for_data_packets(self, port: str) -> None:
...
-
-                        # If the data point is invalid, add it to the invalid fields list:
-                        if not data_point.valid():
-                            if imu_data_packet.invalid_fields is None:
-                                imu_data_packet.invalid_fields = []
-                            imu_data_packet.invalid_fields.append(channel)
-
-                # Put the latest data into the shared queue
+                # Process the packet's data and populate the data packet object.
+                IMU._process_packet_data(packet, imu_data_packet)
```
harshil21, line 207:
```suggestion
                self._process_packet_data(packet, imu_data_packet)
```

airbrakes/mock/mock_imu.py
```diff
@@ -61,22 +65,40 @@ def __init__(
...
-    def _convert_invalid_fields(value):
-        if not value:
-            return None
-        return ast.literal_eval(value)  # Convert string to list
+    def _convert_invalid_fields(value) -> list:
+        """
+        Convert invalid fields to Python objects or None.
+        :param value: The value to convert.
+        :return: The converted value
+        """
```
harshil21, line 81: the comment which said that this is used for converting the string to list should be kept imo

airbrakes/mock/mock_imu.py
```diff
@@ -27,29 +32,28 @@ class MockIMU(IMU):
...
-        start_after_log_buffer: bool = False,
+        log_file_path: Path | None = None,
+        start_after_log_buffer: bool = True,
     ):
         """
         Initializes the object that pretends to be an IMU for testing purposes by reading from a
-            log file.
+        log file.
+
+        We don't call the parent constructor as the IMU class has different parameters, so we
```
harshil21, line 43: i'm still against that btw, it does break LSP. But ok whatever

airbrakes/mock/mock_imu.py
```diff
@@ -61,22 +65,40 @@ def __init__(
...
+
+    @staticmethod
+    def _calculate_start_index(log_file_path: Path) -> int:
+        """
+        Calculate the start index based on log buffer size and time differences.
+        :param log_file_path: The path of the log file to read data from.
+        :return: The index where the log buffer ends.
+        """
+        chunk_size = LOG_BUFFER_SIZE + 1
+        for chunk in pd.read_csv(
```
harshil21, line 92: Let's add a new comment here saying that we process in chunks because it's faster to read a small chunk of the file where we expect the buffer index to be rather than to load the whole file into memory.

airbrakes/mock/mock_imu.py
```diff
@@ -61,22 +65,40 @@ def __init__(
...
+            log_file_path,
+            chunksize=chunk_size,
+            converters={"invalid_fields": MockIMU._convert_invalid_fields},
```
harshil21, line 95:
```suggestion
            converters={"invalid_fields": self._convert_invalid_fields},
```

airbrakes/mock/mock_imu.py
```diff
@@ -116,14 +121,17 @@ def _read_file(
...
+
         # Read the csv, starting from the row after the log buffer, and using only the valid columns
         df = pd.read_csv(
             log_file_path,
-            skiprows=range(1, start_index + 1),
+            skiprows=list(range(1, start_index + 1)),
             engine="c",
             usecols=valid_columns,
-            converters={"invalid_fields": self._convert_invalid_fields},
+            converters={"invalid_fields": MockIMU._convert_invalid_fields},
```
harshil21, line 131:
```suggestion
            converters={"invalid_fields": self._convert_invalid_fields},
```

constants.py
```diff
@@ -45,99 +49,124 @@ class DisplayEndingType(StrEnum):
...
 
-INTEGRATION_TIME_STEP = EST_DATA_PACKET_SAMPLING_RATE
+GRAVITY_METERS_PER_SECOND_SQUARED = 9.798
+"""This is the standard gravity on Earth."""
 
-# This is the standard gravity on Earth, in m/s^2
-GRAVITY = 9.798
+TARGET_ALTITUDE_METERS = 1000
+"""The target altitude in meters that we want the rocket to reach. This is used with our bang-bang
+controller to determine when to extend and retract the airbrakes."""
```
harshil21, line 163: I feel like this should be in the Coast to free fall section, since it's not actually used in apogee_predictor.py

constants.py
```diff
@@ -45,99 +49,124 @@ class DisplayEndingType(StrEnum):
...
+"""The initial guess for the coefficients for curve fit of the acceleration curve."""
+
+APOGEE_PREDICTION_MIN_PACKETS = 10
+"""The minimum number of data packets required to predict the apogee."""
 
-# The uncertainty from the curve fit, below which we will say that our apogee has converged:
-UNCERTAINTY_THRESHOLD = [0.0359, 0.00075]  # [0.0259, 0.00065]
-# UNCERTAINTY_THRESHOLD = [0.1, 0.75]  # [0.0259, 0.00065]
+UNCERTAINTY_THRESHOLD = [0.0359, 0.00075]  # For near quick convergence times, use: [0.1, 0.75]
+"""The uncertainty from the curve fit, below which we will say that our apogee has converged"""
```
harshil21, line 172:
```suggestion
"""The uncertainty from the curve fit, below which we will say that our apogee has converged. This uncertainty
corresponds to being off by +/- 5m."""
```

main.py
```diff
@@ -15,99 +15,105 @@
...
+    # Handle user interrupt gracefully
     except KeyboardInterrupt:
-        if args.mock:
+        if is_mock:
             flight_display.end_sim_interrupted.set()
-    else:  # This is run if we have landed and the program is not interrupted (see state.py)
-        if args.mock:
-            # Usually the mock sim will stop itself due to data exhaustion, so we will actually
-            # hit the condition above, but if the data isn't exhausted, we will stop it here.
+    else:
```
harshil21, line 77: the comment which specified when the else block was executed is important context imo

main.py
harshil21, line 1: I like the refactoring of this a lot. I had plans to do this in #106 anyway, and I'll probably still have to change a lot of it, but it's good to have this here instead in mine.

main.py
```diff
@@ -15,99 +15,105 @@
...
+    run_flight_loop(airbrakes, flight_display, args.mock)
+
+
 if __name__ == "__main__":
-    # Command line args:
+    # Command line args (after these are run, you can press Ctrl+C to exit the program):
     # python main.py -v: Shows the display with much more data
-    # python main.py -m: Runs a simulation on your computer.
+    # python main.py -m: Runs a simulation on your computer
     # python main.py -m -r: Runs a simulation on your computer with the real servo.
```
harshil21, line 113: what's the philosophy behind removing the periods here, but not in the next sentence?

harshil21 (commented):
The cleanup of imu and main.py was definitely needed, nice job. Most of my comments are just nitpicks:

airbrakes/airbrakes.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         :param servo: The servo object that controls the extension of the airbrakes. This can be a
         real servo or a mock servo.
         :param imu: The IMU object that reads data from the rocket's IMU. This can be a real IMU or
-            a mock IMU.
+        a mock IMU.
```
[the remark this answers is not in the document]
JacksonElia: cope about it for now and make it something in ruff, idc either way, I just want it to be consistent

airbrakes/data_handling/logger.py
```diff
@@ -249,4 +246,4 @@ def _logging_loop(self) -> None:
                 # If the message is the stop signal, break out of the loop
                 if message_fields == STOP_SIGNAL:
                     break
-                writer.writerow(self._truncate_floats(message_fields))
+                writer.writerow(Logger._truncate_floats(message_fields))
```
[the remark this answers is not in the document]
JacksonElia: yeah we can, but I like that it immediately makes it explicit that we are calling a static method. Additionally, if one static method is calling another static method, it doesn't have access to self, and so you would have to do this anyways.

airbrakes/data_handling/logger.py
```diff
@@ -220,19 +230,6 @@ def _prepare_log_dict(
 
         return logged_data_packets
 
-    # ------------------------------- RUN IN A SEPARATE PROCESS -----------------------------------
```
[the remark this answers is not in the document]
JacksonElia: I think the docstring shows it well enough and its weird doing it for just 1 method (not including the static method). I don't care about this thaaat much so if you really want it, I'll add it back.

airbrakes/hardware/imu.py
```diff
@@ -66,6 +69,66 @@ def is_running(self) -> bool:
         """
         return self._running.value
 
+    @staticmethod
+    def _initialize_packet(packet) -> IMUDataPacket | None:
```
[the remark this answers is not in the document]
JacksonElia: good catch

constants.py
JacksonElia: You made me think about this for a minute and I think you're right

main.py
JacksonElia: yeah its a good comment

main.py
JacksonElia: I forgot this line lol

airbrakes/hardware/imu.py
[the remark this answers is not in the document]
JacksonElia: wait I remembered why I deleted it -- for some reason the method being static wasn't playing nicely with the mscl library and was failing our unit tests. If you can fix this, please do. All of the static methods in this class need typing

airbrakes/data_handling/logger.py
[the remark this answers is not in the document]
harshil21: yeah things need to be consistent with the other processes. Also, nobody reads the docstrings at first glance.

airbrakes/hardware/imu.py
```diff
@@ -66,6 +69,66 @@ def is_running(self) -> bool:
...
+    @staticmethod
+    def _process_data_point(data_point, channel: str, imu_data_packet) -> None:
+        """
+        Process an individual data point and set its value in the data packet object.
+        :param data_point: The IMU data point containing the measurement.
+        :param channel: The channel name of the data point.
+        :param imu_data_packet: The data packet object to update.
+        """
+        # Handle special channels that represent quaternion data.
+        if channel in {"estAttitudeUncertQuaternion", "estOrientQuaternion"}:
```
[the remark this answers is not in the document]
JacksonElia: lol it was breaking for me for some reason, plus this is cleaner

airbrakes/mock/mock_imu.py
JacksonElia: Lol this one is on you. When I was first implementing the MockIMU, I was going to make an abstract class to just represent an IMU, but you said to just have the mock imu inherit from imu

harshil21: but an abstract class implies that MockIMU is different from IMU and all its methods would be implemented differently. We want to make sure MockIMU is as close as possible to the IMU class though, so abstract classes isn't the way

airbrakes/hardware/imu.py
```diff
@@ -123,69 +184,27 @@ def _query_imu_for_data_packets(self, port: str) -> None:
...
-            # During IMU configuration (outside of this code), we set the sampling rate of the IMU
-            # as 1ms for RawDataPackets, and 2ms for EstimatedDataPackets.
-            # So we use a timeout of 10ms which should be more
-            # than enough. If the timeout is hit, the function will return an empty list.
-
-            packets: mscl.MipDataPackets = node.getDataPackets(timeout)
-            # Every loop iteration we get the latest data in form of packets from the imu
+            # Reconnect to the IMU at the start of each loop iteration.
+            connection = mscl.Connection.Serial(port)
+            node = mscl.InertialNode(connection)
```
harshil21, line 194: I have no idea what you said since you deleted it, but why reconnect?

airbrakes/hardware/imu.py
```diff
@@ -123,69 +184,27 @@ def _query_imu_for_data_packets(self, port: str) -> None:
             self._fetch_data_loop(port)
 
     def _fetch_data_loop(self, port: str) -> None:
```
harshil21, line 186: also deleted - move port to self variable?

harshil21 (commented):
good thing I remembered what I reviewed

airbrakes/hardware/imu.py
JacksonElia: it was me being dumb basically

airbrakes/hardware/imu.py
JacksonElia: no

airbrakes/mock/mock_imu.py
JacksonElia: I think the "correct" way to do this would be making a parent class both MockIMU and IMU extend. This parent class would contain as much functionality as it could, and both MockIMU and IMU would only have their own specific constructors and implementations of `_fetch_data_loop`

harshil21: I'm fine with that, maybe call it `BaseIMU`?

JacksonElia: Yeah something like that.

airbrakes/hardware/imu.py
harshil21: yeah that's why I was putting those comment separators, even in the init. Since we have long running processes, and many methods, an object oriented approach is still the best imo.

harshil21 (approved):

PR #113, opened 2024-12-03
Adding Fields and Missing Data to csv's

wlsanderson:
This was going to be added in #107, but that is being closed for now. I wanted to still add these modified csv's though. Purple Nurple has 4 columns added for quaternion uncertainty, and the missing quaternion data for interest launch was added by calculating them from gyroscope integration. See the last comment and attached images of #107 for the accuracy of the calculated quaternions. Interest launch still has discontinuities in quaternion data due to the IMU, but this is not something that can be accounted for.

### DIFF
--- a/launch_data/README.md
+++ b/launch_data/README.md
@@ -2,13 +2,13 @@ This folder contains the launch data of actual previous flights. These are used 
 
 ## Data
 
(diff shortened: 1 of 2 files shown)


harshil21 (approved):
LGTM

PR #118, opened 2025-01-01
Use acceleration spike to switch to `LandedState`

### LINKED ISSUE #114: `LandedState` should be detected based on acceleration spike
Genesis Flight 1 had trouble switching to `LandedState`, because the velocity calculated from the gyro was off by > 20m/s due to drift. Genesis Flight 2 lucked out and the velocity was mostly accurate and consistent with velocity from the altitude data. 

Accurately detecting landing is critical because we have to stop logging. The only safe and sensible way to detect landing is based on the acceleration spike (along with height lower than `LANDED_ALTITUDE` =~ 10m). 

The same point can be made for detecting MotorBurn, where we currently switch to MotorBurn a bit too slowly, e.g. it takes 50 est packets to switch whereas we could've done it in 6 est packets (where the accel was -30m/s^2). So we were about 10x slower to switch states.

harshil21:
Closes #114.

To be merged after #117.

All launch data showed a spike of vertical acceleration of well over 100m/s^2 upon impact. I chose a safe value of 70 and above.


airbrakes/data_handling/data_processor.py
```diff
@@ -85,6 +85,11 @@ def max_vertical_velocity(self) -> float:
         """The maximum vertical velocity the rocket has attained during the flight, in m/s."""
         return float(self._max_vertical_velocity)
 
+    @property
```
wlsanderson, line 88: See #125

tests/test_integration.py
```diff
@@ -194,14 +183,15 @@ def test_update(
...
         assert all(ext == ServoExtension.MIN_EXTENSION for ext in free_fall_state.extensions)
 
+        # Check landed state values:
+        landed_state = states_dict.LandedState
         if launch_name != "purple_launch":
-            # Generated data for simulated landing for interest launcH:
-            landed_state = states_dict.LandedState
-            assert landed_state.min_velocity >= -LANDED_SPEED_METERS_PER_SECOND
+            # Generated data for simulated landing for interest launch:
+            assert landed_state.min_velocity >= -15.0  # high velocity impact
```
wlsanderson, line 190: Shouldn't this be checking acceleration spike, not velocity?

harshil21: good point. I'll adjust the integration test for that

wlsanderson (approved):

Issue #111, opened 2024-12-02
Refactor IMU: `BaseIMU` -> [`MockIMU`, `IMU`]


harshil21:
Split the IMU and MockIMU class into a BaseIMU

_Originally posted by @JacksonElia in https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/109#discussion_r1866557848_

unclesam79:
Created PR #122 for this.

PR #122, opened 2025-01-04
#111 Refactor IMU: BaseIMU ->  [MockIMU, IMU]

unclesam79:


airbrakes/hardware/base_imu.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def is_running(self) -> bool:
+        """
+        Returns whether the process fetching data from the IMU is running.
+        :return: True if the process is running, False otherwise
+        """
+        return self._running.value
+
+    @is_running.setter
+    def is_running(self, value: bool) -> None:
```
harshil21, line 89: I understand the appeal of using a setter here, however, this setter property must not exist since we don't want to signal the other classes that this is modifiable. It can only be controlled by `stop()`.

airbrakes/hardware/imu.py
harshil21, line 1: can you replace the use of `self._running.value` with `self.is_running` (in the loop at line 132)

airbrakes/mock/mock_imu.py
harshil21, line 1: same thing here, we can use the property

harshil21 (changes requested):
test failures seem unrelated (I need to still run locally and check). Tests don't really need to be added/modified since nothing has changed functionally.

pre-commit needs to be fixed tho

harshil21 (approved):
LGTM

PR #121, opened 2025-01-02
Use `uv` to streamline setup and update README

harshil21:
Closes #120.

See the magic for yourself! On any machine, install uv, and just run:

`uvx --from git+https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2.git@use-uv mock`

To be merged after #117.

PR #123, opened 2025-01-06
Implement Rotary Encoder

harshil21:
Implements the rotary encoder into the code. 

As of right now, there is only a testing script which prints a super basic GUI to control the servo and print the encoder position. 

Run with: `uv run scripts/run_servo_and_encoder.py`.

merge after #121


wlsanderson:
tested on the pi. Pins changed to 23 and 24, and I changed the code to use the built-in way to check for encoder updates, rather than using a time.sleep. The time.sleep causes the encoder to skip a point when rotating very quickly, which produces an inaccurate measurement. The built-in method can still skip points occasionally, but it is smoother. This code isn't really clean though and I can probably go back and clean it up later.

PR #126, opened 2025-01-09
Automatically install `mscl` via pip

harshil21:
I did a thing and packaged mscl - https://github.com/harshil21/python-mscl so it's easily installable


harshil21:
Tested this on the Pi just now (`uv run real`). Works flawlessly.

JacksonElia (approved):
lgtm

PR #127, opened 2025-01-09
Rotary Encoder script

harshil21:
Github closed #123 automatically so here's it again


JacksonElia (approved):
This looks good, lets make a separate pr off of #100 adding a servo data packet and having the rotation being logged

harshil21:
needed to merge this because this has the buzzer script which runs on hotspot connection, and thus needs to be on main so it works all the time

PR #128, opened 2025-01-09
Renaming "Mock Sim" to "Mock Replay"

wlsanderson:
With branch #76 creating an actual flight simulation, I wanted to remove mentions of the word "simulation" and "sim" in our code base, and instead re-label our mock sim as "mock replay" instead.


JacksonElia (approved):
lgtm

PR #129, opened 2025-01-13
Using acceleration to switch to motor burn state

wlsanderson:
With our current state change, it takes around 50 packets or so to switch into motor burn, but this shortens it down to just 3 or 4. I think 25 m/s^2 is well-within the safe range and won't be triggered prematurely while on the pad. Even if it does, we now have the means of checking that with the display, and can reset it.

I kept the other conditions for velocity and height in there, but I think that we should consider removing at least height.

### DIFF
--- a/airbrakes/state.py
+++ b/airbrakes/state.py
@@ -9,7 +9,6 @@ from airbrakes.constants import (
     MAX_FREE_FALL_SECONDS,
     MAX_VELOCITY_THRESHOLD,
     TAKEOFF_ACCEL_METERS_PER_SECOND_SQUARED,
-    TAKEOFF_HEIGHT_METERS,
     TAKEOFF_VELOCITY_METERS_PER_SECOND,
     TARGET_ALTITUDE_METERS,
 )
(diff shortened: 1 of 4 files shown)


wlsanderson:
I ended up just removing the height condition

harshil21 (approved):

airbrakes/state.py
```diff
@@ -81,16 +81,16 @@ def update(self):
...
+        # launched.
 
         data = self.context.data_processor
 
         if data.vertical_velocity > TAKEOFF_VELOCITY_METERS_PER_SECOND:
             self.next_state()
             return
 
-        if data.current_altitude > TAKEOFF_HEIGHT_METERS:
+        if data.average_vertical_acceleration > TAKEOFF_ACCEL_METERS_PER_SECOND_SQUARED:
```
JacksonElia, line 93: this is a nit, but because this usually gets triggered before the velocity, I would put this first.

JacksonElia (approved):
lgtm

PR #136, opened 2025-01-18
Use arm64 runners on Github Actions

harshil21:
Now also run the tests on arm64, which is the CPU architecture the raspberry pi uses.


harshil21:
We could also artificially limit the CPU in github actions to more closely match the raspberry pi's compute power, but I don't think that's worth the much slower execution time.

JacksonElia (approved):

PR #138, opened 2025-01-20
Fix bug, lots of tests and some restructuring

### LINKED ISSUE #137: Rare: Program does not cleanly stop if we landed and hit the log buffer in one case
Something more alarming after fixing the display attribute error on #134: If you run `uv run mock -v -f -p launch_data/genesis_launch_1.csv` there, everything works, but sometimes randomly the mock has a delayed stop, because there is an exception saying that the queue is Full in MockIMU.

Noticeably, this only happens for genesis launch 1, and only on the mock. Here's why:

1. That launch has generated data packets for landing.
2. After #118, we switch to LandedState and still have many packets left to read in the file.
3. The buffer accumulates and once we hit 5000, we command to stop the program.
4. Now, if our imu queue size is full at that instant (that's 15 packets when running the mock), then we deadlock on the .put() since we never hit the condition in the loop which checks whether we should stop or not.

Why would it very likely not happen in a real flight?
- Because our max queue size is 100k packets, which gives more than enough time to stop everything without any deadlocks.

Suggested fix:
- Fetch any remaining packets in BaseIMU.stop(), like how we used to do. It will need to non-blocking, otherwise that's another potential deadlock situation. So `get_imu_data_packets()` must accept a new parameter.

harshil21:
Closes #137 

See commit messages for the more detailed changes.


JacksonElia:
while I think we needed to organize stuff better than we were, and I like how you structured stuff, idk if telemetry is the right word. In an aerospace context doesn't telemetry mean collecting data and transmitting it to the ground?

harshil21:
yeah telemetry involves transmission too, but a part of it also means handling, measuring and processing the data, which is what we do. It's not perfect, but it does sound better than just `data_handling` imo. If you have another suggestion which is one word long and more accurately describes it, i'm down for it

JacksonElia:
i think we should make it 3 words

JacksonElia:
can we merge this after the launch? it just modifies like every file and i don't feel liek it

harshil21:
> can we merge this after the launch? it just modifies like every file and i don't feel liek it

there's an important bug fix on here tho, and more tests, which is also important. If you want I can revert back the folder name change for now

wlsanderson:
I think these are good changes. idrc about the telemetry name change, im fine either way but I don't really see a reason to change it. Otherwise, lgtm

airbrakes/hardware/servo.py
```diff
@@ -41,16 +40,16 @@ def __init__(self, gpio_pin_number: int, pin_factory=None) -> None:
...
+            # We use a MockFactory for testing, which simulates the GPIO pins:
             gpiozero.Device.pin_factory = pin_factory
-            # When running the mock, we don't have an encoder
-            self.encoder = None
 
-        self.servo = gpiozero.Servo(gpio_pin_number, initial_value=ServoExtension.MIN_NO_BUZZ.value)
+        self.servo = gpiozero.Servo(gpio_pin_number, initial_value=self.current_extension.value)
+        # max_steps=0 indicates that the encoder's `value` property will never change. We will
+        # only use the integer value, which is the `steps` property.
+        self.encoder = gpiozero.RotaryEncoder(ENCODER_PIN_A, ENCODER_PIN_B, max_steps=0)
```
JacksonElia, line 52: wait why did you move this out of the if statement? I had it there so we could do it in the mock

harshil21: setting the gpio.Device.pin_factory thingy automatically makes it work for the mock as well, like how it is for the servo

JacksonElia:
Some of these tests fail when I run them on the pi
image

JacksonElia:
Also other things I noticed from testing that aren't maybe originally caused by this branch, but still need to be fixed (I ran the main script like we do when we launch):
1. The velocity kept decreasing even though nothing was happening, so it was sitting on the table and said it was going at -60 m/s
2. Something about the camera process printed (maybe a warning) and it fucked up the whole display

harshil21:
> Some of these tests fail when I run them on the pi image

yeah I'm aware. I suspect it might just be due to the slow processing power things just take longer to run? All the failing tests are related to stopping right?

harshil21:
> Also other things I noticed from testing that aren't maybe originally caused by this branch, but still need to be fixed (I ran the main script like we do when we launch):
> 
> 1. The velocity kept decreasing even though nothing was happening, so it was sitting on the table and said it was going at -60 m/s

Maybe due to invalid quaternions? 

> 2. Something about the camera process printed (maybe a warning) and it fucked up the whole display

What did it say?

JacksonElia:
I'll test again, but it didn't happen on main, and I'll send a screen shot.

JacksonElia:
Okay so this isn't unique to just this branch but maybe its random cause it happened to me on main too
image

JacksonElia:
camera problem happens everytime on that branch but I think the velocity thing I saw was just bad quats

harshil21:
why aren't using `uv run mock` tho. Run with `uv run mock -d` so the display isn't interfering. Also run `uv sync --all-groups` before running the mock

JacksonElia:
I'm not running a mock I'm running it on the pi

harshil21:
oh then I mean `uv run real -d`

JacksonElia:
image

harshil21:
I didn't get that yesterday. Does that also happen on main? Run `uv sync --all-groups`, then the `uv run real -d` command?

JacksonElia:
Yeah I tried and it still happened. Also the camera error is actually an index out of bounds, just cause no camera is plugged in. Would be nice to have something to suppress the error message.

harshil21:
Ok the camera thing can be wrapped in a try except or similar. The uv thing is a bit concerning though, can you try using `rm` to delete the file?

JacksonElia:
no u got it I dont wanna touch ts

airbrakes/hardware/servo.py
JacksonElia: oh good find

JacksonElia (approved):
Lets test encoder logging to be safe

PR #139, opened 2025-01-21
Revert switch to Motorburn from accel

JacksonElia:
I reverted the motor burn change so now it only uses velocity to detect motor burn. This was after I realized a very mild jolt on the table triggered it.

Also I wrapped the camera in a try catch


harshil21:
the acceleration only exceeds 35 m/s^2 when you jerk the rocket downwards (so gravity is added). It's not really an issue when the rocket is on the rail, since it will only go up not down :) 

that being said, if you still want to remove this, you need to add back the altitude based state change again (see commit history for tests + changes)

JacksonElia:
me and will talked about this and I agreed with him that altitude was totally worthless. But also what you said about only going down is really true when it comes to a jolt because it goes up then down (idk will explained it better)

PR #141, opened 2025-01-23
Refactor argument parsing

harshil21:
Refactors argument parsing to be more concise, explicit, and error free.

1. Unifies the legacy way of starting the script and the `uv` way.
2. Uses built-in mechanism for mutually exclusive arguments
3. Splits the argument parsing into 3 separate subparsers.
4. Better and more exhaustive tests
5. Makes SimIMU work with `-f`.

Merge after legacy launch. Or before (but then I would have to rebase)

### DIFF
--- a/airbrakes/main.py
+++ b/airbrakes/main.py
@@ -2,6 +2,7 @@
 and run the main loop."""
 
 import argparse
+import sys
 import time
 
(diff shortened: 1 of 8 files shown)


wlsanderson:
can we do this after launch lol

Issue #115, opened 2024-12-12
Add `pytest-benchmark` to the test suite


harshil21:
We should consider adding [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/usage.html) to the test suite. This will allow us to keep track of the performance of the code over time.

Additionally there should be another Github action to compare the latest benchmark run on the head of a branch and compare it against `main`. We can export the .png as artifacts to allow for easy viewing in the browser itself.

PR #146, opened 2025-01-28
Optimize Reading IMU Data

Graicc:
In the legacy launch, the IMU data was significantly behind real time. This was due to us not processing IMU data packets fast enough, which cased us to be processing a backlog of ground packets while in the air. This PR improves the performance of the IMU reading. However, we will still need to lower the frequency to 500hz for the python code to be fast enough.

TODO:
- [x] Fix unused channels
- [ ] ~Implement 1000hz -> 500hz change in code~
- [ ] ~Use constants (from MSCL) for field and qualifier~
- [x] Remove methods that have been inlined
- [x] Clean up
- [x] Log the number of fetched packets from the IMU
- [x] Tests and benchmarks


harshil21:
Just tried printing fetched packets from IMU at 1000Hz raw and 500Hz estimated on this branch and on main. On main the number keeps increasing to 10k packets after 10-15 seconds. On this branch, even at those frequencies, the number of packets don't exceed 50.

However, if I keep the script running, and start typing in VSCode, the packets spike quite a bit. So we definitely need to assign a higher priority to our process compared to other processes. 

After I implement #147 and #115, we should never have this issue again.

Thanks @Graicc!

harshil21:
Alright I think this branch is good to be merged. There is now an benchmark workflow too. The results of which are uploaded to github pages - https://ncsu-high-powered-rocketry-club.github.io/AirbrakesV2/bench/imu-fix/index.html. The url has the branch name in it, so you can substitute the name in for whatever branch you want to see.

Unfortunately the workflow does not provide a way to compare the run against another branch like `main` - https://github.com/benchmark-action/github-action-benchmark/issues/267. Or maybe there is a way to hack around it? Anyway, the graph is a good indicator of regressions, so I don't mind just using that. The PR comment also works well.

Not doing the IMU polling freq change in code since it seems too obscure and it didn't work with raw data for some reason. I don't know what @Graicc meant by "Fix unused channels", I believe in sensor connect we already only select all the channels we want.

airbrakes/hardware/imu.py
```diff
@@ -64,99 +70,117 @@ def __init__(self, port: str) -> None:
...
-                # Process the packet's data and populate the data packet object.
-                IMU._process_packet_data(packet, imu_data_packet)
-
-                # Add the processor data packet to the shared queue.
-                self._data_queue.put(imu_data_packet)
+                # Iterate through each data point in the packet.
+                for data_point in packet.data():
+                    # Extract the channel name of the data point.
+                    match data_point.field(), data_point.qualifier():
+                        case 32775, 1:
```
harshil21, line 111: todo: test if this is faster than just using `if-else`? If it isn't, write a comment saying by how much it is slower by.

airbrakes/hardware/imu.py
```diff
@@ -64,99 +70,117 @@ def __init__(self, port: str) -> None:
...
-
-    def _fetch_data_loop(self, port: str) -> None:
+    def _fetch_data_loop(self, port: str) -> None:  # pragma: no cover
         """
         Continuously fetch data packets from the IMU and process them.
         :param port: The serial port to connect to the IMU.
         """
+        # Set the process priority really high, as we want to get the data from the IMU as fast as
+        # possible:
+        set_process_priority(-15)
```
harshil21, line 80: before/after merge, do this change on the pi - https://www.perplexity.ai/search/how-do-i-run-a-script-as-sudo-cSdzzQURTV2OjGnO5CtgXw#0

airbrakes/hardware/imu.py
```diff
@@ -64,99 +70,117 @@ def __init__(self, port: str) -> None:
...
-                if imu_data_packet is None:
-                    # Skip unrecognized packets.
+                # Extract the timestamp from the packet.
+                timestamp = packet.collectedTimestamp().nanoseconds()
+
+                # Initialize packet with the timestamp, determines if the packet is raw or estimated
+                if packet.descriptorSet() == ESTIMATED_DESCRIPTOR_SET:
+                    imu_data_packet = EstimatedDataPacket(timestamp)
+                elif packet.descriptorSet() == RAW_DESCRIPTOR_SET:
+                    imu_data_packet = RawDataPacket(timestamp)
```
harshil21, line 103:
```suggestion
                if packet.descriptorSet() == RAW_DESCRIPTOR_SET:
                    imu_data_packet = RawDataPacket(timestamp)
                elif packet.descriptorSet() == ESTIMATED_DESCRIPTOR_SET:
                    imu_data_packet = EstimatedDataPacket(timestamp)
```
micro-optimization: since raw freq is 2x est, this shortens the path the code has to take? Not sure if the newer python versions bytecode already optimize for this but worth doing anyway

airbrakes/hardware/camera.py
```diff
@@ -50,44 +50,42 @@ def start_recording(self):
...
+
+        # Start recording with the buffer. This operation is non-blocking.
+        camera.start_recording(encoder, output)
+
+        # TODO: this is ass
+        # Check if motor burn has started, if it has, we can stop buffering and start saving
+        # the video. This way we get a few seconds of video before liftoff too. Otherwise, just
+        # sleep and wait.
+        while not self.motor_burn_started.is_set():
+            # Unfortunately, an `multiprocessing.Event.wait()` doesn't seem to work, hence this
```
harshil21, line 80: todo: test with event.wait() again

airbrakes/hardware/base_imu.py
```diff
@@ -39,6 +40,15 @@ def __init__(self, data_fetch_process: Process, data_queue: Queue) -> None:
         self._data_queue = data_queue
         # Makes a boolean value that is shared between processes
         self._running = Value("b", False)
+        self._fetched_imu_packets = Value("i", 0)
```
harshil21, line 43: todo: test these mp.Value's with lock=False

airbrakes/hardware/imu.py
harshil21: so if-elif-else statements are faster. Benchmarked this:

- Current implementation with `match` statements: 7546 ns.
- With if-elif-else: 4188 ns. 

I ran it for 15 seconds, so it should be a good representation of the performance difference

airbrakes/hardware/camera.py
harshil21: it works, yay

airbrakes/airbrakes.py
```diff
@@ -83,20 +83,19 @@ def __init__(
...
         self.servo_data_packet: ServoDataPacket | None = None
         self.last_apogee_predictor_packet = ApogeePredictorDataPacket(0, 0, 0, 0, 0)
 
-        self._update_count: int = 1
-
     def start(self) -> None:
         """
         Starts the IMU and logger processes. This is called before the main while loop starts.
         """
+        set_process_priority(-5)  # Higher than normal priority
```
JacksonElia, line 98: make a constant for the priorities

airbrakes/telemetry/packets/context_data_packet.py
```diff
@@ -26,3 +27,12 @@ class ContextDataPacket(msgspec.Struct):
     apogee_predictor_queue_size: int
     """The number of apogee predictor data packets in the apogee predictor queue,
     waiting to be fetched by the main process."""
+
+    fetched_imu_packets: int
```
JacksonElia, line 31: can we make better names for this and `fetched_packets_in_main`, at least something like `fetched_imu_packets_in_main`

airbrakes/hardware/imu.py
```diff
@@ -64,99 +70,165 @@ def __init__(self, port: str) -> None:
...
-
-    def _fetch_data_loop(self, port: str) -> None:
+    def _fetch_data_loop(self, port: str) -> None:  # pragma: no cover
         """
         Continuously fetch data packets from the IMU and process them.
         :param port: The serial port to connect to the IMU.
         """
+        # Set the process priority really high, as we want to get the data from the IMU as fast as
+        # possible:
+        set_process_priority(-15)
```
JacksonElia, line 80: constant

airbrakes/hardware/imu.py
```diff
@@ -64,99 +70,165 @@ def __init__(self, port: str) -> None:
...
+        # the way as it is for performance reasons. Some of the optimizations implemented include
+        # - using no functions inside the loop (this is typically 2x faster per packet)
+        # - if-elif statements instead of a `match` / `setattr` / `hasattr` (2x-5x faster / packet)
+        # - using 2 inner loops depending on the type of packet, this reduces the number of `if`
+        # checks the interpreter has to do
+        # - Using integers for comparison instead of strings (O(1) vs O(n) complexity)
+        # - Checking for raw data packets first, since that is 2x the frequency of estimated packets
+        # - Using msgspec to serialize and deserialize the packets, which is faster than pickle
+        # - High priority for the process
         while self.is_running:
```
JacksonElia, line 98: we need to make constants for this entire thing, if I feel like it I might do that

JacksonElia: also we really should absolutely make sure each field is being set properly.

airbrakes/mock/display.py
```diff
@@ -143,7 +147,16 @@ def sound_alarm_if_imu_is_having_issues(self) -> None:
...
+        if self._airbrakes.imu.fetched_imu_packets > 50:
+            imu_queue_backlog = True
+
+        # if (
+        #     self._pitch_at_startup
+        #     and abs(self._airbrakes.data_processor.average_pitch - self._pitch_at_startup) > 1
+        # ):
+        #     pitch_drift = True
+
+        if has_invalid_fields or has_negative_velocity or imu_queue_backlog or pitch_drift:
```
JacksonElia, line 159: can we make it so that if any of these things happens, it turns all font red? I want it to be incredibly clear and obvious.

airbrakes/utils.py
```diff
@@ -22,10 +28,26 @@ def get_always_list(self, *args, **kwargs) -> list:
     as the multiprocessing.Queue doesn't have a `get_many` method"""
     fetched = self.get(*args, **kwargs)
     if isinstance(fetched, list):
+        if isinstance(fetched[0], LoggerDataPacket):
```
JacksonElia, line 31: isn't isinstance slower than type?

harshil21: it's only for Windows which isn't running on the pi so I don't really care tbh

airbrakes/mock/display.py
harshil21: is the alarming sound it produces not enough? tbh in the sunlight red would be even harder to see ironically

airbrakes/telemetry/packets/context_data_packet.py
harshil21: I asked @wlsanderson and he said the name is too long but I know you like long names and I'm indifferent but agree that we need a better / more descriptive name

harshil21:
libcamera was updated, so we need to check the whole camera integration again + auto white balancing. Should be good to merge after that.

airbrakes/telemetry/packets/context_data_packet.py
JacksonElia: okay lets go ahead and add it, its only 4 more characters

airbrakes/constants.py
```diff
@@ -3,6 +3,16 @@
...
 from pathlib import Path
 
+# -------------------------------------------------------
+# Main Configuration
+# -------------------------------------------------------
+
+MAIN_PROCESS_PRIORITY = -11
+"""The priority of the Main process. This is a really high priority so the OS knows to give it
+priority over other processes. This is because we want to make sure we don't want to stay behind
+on processing the packets from the IMU."""
```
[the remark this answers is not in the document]
harshil21: how should it be different then

harshil21:
There's been a mysterious performance regression in the last few commits. It was even noticeable on my end, slowing down `uv run mock -v -f` from ~ 2.54 seconds (from what I remember) to ~3.54 seconds. A whole second. This was even noticeable in the benchmarks - https://ncsu-high-powered-rocketry-club.github.io/AirbrakesV2/bench/imu-fix/index.html, failing the CI on https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/146/commits/030256066cd07d3d36e5eb31861f6ae883e356ff

I have no idea why this happened, even if I checkout https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/146/commits/26040b3bcd7ccbbd03346ae8e2b606c6951bd23a and run the mock from there, the regression is still present. Moreover the regression is also present in #154.

So that rules out the renaming. I can also rule out the pitch related changes.

wlsanderson (approved):
lgtm

harshil21:
latest commit uses `lock=False`, which gives a performance boost (~90% on every read/write), since we are reading and writing in separate processes, and thus don't need a Lock. On modern architectures, the read and write of simple types like ints, and bools are atomic, so there's no risk of data corruption.

wlsanderson (approved):

PR #154, opened 2025-02-10
Initial Cython implementation 

harshil21:
Adds Cython to the build process to compile the .py files and try to speed the IMU even more.

Currently, every .py file is compiled, and I didn't see any noticeable performance improvements right off. In fact it seems to be slightly slower if anything (checked by looking at how long it takes to run a mock with `-f`). Which is kinda sad since Cython's documentation said you'd get a 30-60% speedup by simply compiling and not changing anything.

Some notes on the changes so far: 
- Cython does not support match statements yet, so I had to convert them to `if-else`. https://github.com/cython/cython/issues/4029
- Cython caught some type bugs, but most of our code is typed correctly. 
- I am compiling with `boundscheck=False`, which means that all accesses of a list/string/tuple via `[]` must not raise `IndexError`, otherwise the code will segfault/crash.

Performance changes we could do:

- Half of our loops are still going to be done in python since they are using python objects. E.g. `imu_data_packets` contains both Raw and Est packets, so we would literally have to check the type every time in the loop. So unless we change the architecture significantly, there might not be much we can do.
- We can tell cython to inline some of our functions, would be useful in the IMU loop for example.
- The numpy part can be optimized a little more, we can directly access the memoryview and do a little bit of pre-allocation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ out/
 logs/
 
 build/
-AirbrakesV2.egg-info/
+airbrakes.egg-info/
 .venv/
 
 test_logs/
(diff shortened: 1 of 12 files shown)


harshil21:
Was closed because we got a Pi 5 and Cythonizing stuff is probably too much work (AirbrakesV3) and not worth shaving off a few milliseconds.

PR #156, opened 2025-02-20
Use 2 servos, new board, new logic

harshil21:
Changes logic to control two servos at once using the adafruit circuit python library. 

Also rearranges some files around to a new package called `interfaces`.


harshil21:
Performance benchmark is failing because the frequency for Raw packets were turned to 500 Hz

JacksonElia (approved):

PR #163, opened 2025-02-24
Integrating vertical velocity for altitude during air brakes extension

wlsanderson:
Due to the pressure change that air brakes extension causes, I think integrating the vertical velocity during the extension period is a good way to keep altitude readings mostly accurate. There is error that comes with integrating three times (integrating gyro for rotating acceleration, integrating for velocity, integrating for altitude), so I might look into filtering algorithms if simply integrating is not accurate enough. 

We are also going to do some physical testing to see if the pressure issue can be prevented entirely with changes to the AVAB, so this branch is not to be merged until that testing is complete

### DIFF
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -196,6 +196,8 @@ class AirbrakesContext:
         # Because the IMUDataProcessor only uses Estimated Data Packets to create Processor Data
         # Packets, we only update the apogee predictor when Estimated Data Packets are ready.
         if self.est_data_packets:
+            # if not self.last_apogee_predictor_packet.predicted_apogee:
+            #     self.data_processor.filter_vertical_velocities()
             self.apogee_predictor.update(self.processor_data_packets)
 
     def generate_data_packets(self) -> None:
--- a/airbrakes/airbrakes.py
+++ b/airbrakes/airbrakes.py
@@ -179,12 +179,14 @@ class AirbrakesContext:
         Extends the air brakes to the maximum extension.
         """
         self.servo.set_extended()
+        self.data_processor.use_integrated_altitude = True
 
     def retract_airbrakes(self) -> None:
         """
         Retracts the air brakes to the minimum extension.
         """
         self.servo.set_retracted()
+        self.data_processor.use_integrated_altitude = False
 
     def predict_apogee(self) -> None:
         """
--- a/airbrakes/telemetry/data_processor.py
+++ b/airbrakes/telemetry/data_processor.py
@@ -24,16 +24,20 @@ class IMUDataProcessor:
         "_current_altitudes",
         "_current_orientation_quaternions",
         "_data_packets",
+        "_filtered_velocity",
         "_initial_altitude",
         "_integrated_altitudes",
         "_last_data_packet",
-        "_last_valid_altitude",
         "_max_altitude",
         "_max_vertical_velocity",
+        "_pre_extension_altitude",
         "_pressure_altitudes",
+        "_previous_altitude",
+        "_previous_altitude_timestamp",
         "_previous_vertical_velocity",
         "_rotated_accelerations",
         "_time_differences",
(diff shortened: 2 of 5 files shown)


wlsanderson:
I still need to check with pelicanator, but from testing with genesis launch and interest launch, deploying airbrakes immediately after convergence, the error on pressure altitude vs integrating altitude was small... but significant enough to where it could be an issue.
- Genesis: 0.5% error, integrated for 6.8 seconds
- Interest Launch: 1.3% error, integrated for 15.3 seconds.
Apogee was off on interest launch by 24.31 meters. This would be less on Pelicanator, I think the error is still significant. I have no mock to verify with at the moment, but my guess is around 5-8 meters error.

One thing that I noticed a few months ago, before Genesis, was that integrating to get velocity seemed to undershoot, specifically during motor burn. After motor burn, integration was fine. This is most likely due to the IMU having a very high error at high g's.

A jank solution to fix the integration error would be to calculate this velocity error in motor burn, and just multiply the vertical velocity by some amount to "correct" the velocity during motor burn. This would cause integration during air brakes extension to be a little more accurate.

The proper way to fix this issue is using an EKF filter. I don't think filtering the altitude during air brakes extension is a good idea, but filtering the velocity (integrating acceleration, differentiating altitude) would be the safest. This would kill two birds with one stone, fixing the velocity error due to motor burn integration and also getting a more accurate altitude integration during air brakes extension.

harshil21:
Any probable solution would have to be thoroughly plotted and compared for all flights.. I do want to keep the EKF filter as a last resort thing though. 

One other way would be to dynamically calculate the pressure drop dP/dt during airbrakes deployment, add that back to the pressure, and then use this https://en.wikipedia.org/wiki/Pressure_altitude formula to get the altitude back. I did that for a couple of values and got within ~0.75m of what the IMU reported. 

We should plot the "true" altitude line by interpolating for the altitude when airbrakes were deployed, and then we can plot all these solutions and compare it to that "true" line.

harshil21:
image
Also we've not really talked about this, but the convergence parameters for interest launch isn't good enough it seems, our first predicted apogee and actual apogee is quite far off..

JacksonElia:
Yeah tbh with how much we have at stake, I don't think we should risk doing something like this for this year. Ideally we can fix it with some physical changes to the rocket. If not, airbrakes still worked and had us really close to apogee. Worst case we can just set the target apogee in the code like 30-60 ft higher to account for any anticipated problem.

wlsanderson:
The last commit in this branch is definitely getting reverted, I was just trying some stuff. But the "worst case" is  we just integrate vertical velocity with no further changes. And that would mean we are off by 5-8 meters. There are some other options we can look into, but doing nothing and hoping for the best is going to be worse

harshil21:
> I would almost rather just change our target apogee.

that would be the most dangerous thing to do imo. Much better to just do nothing at that point. 

The complimentary filter seemed to be decent, I don't believe it was tested on pelicanator launch yet though.. I don't think we should only use one metric - it needs to be combined with pressure altitude at the very least.

wlsanderson:
The issue with physical changes is that we are certain we can't make physical changes that will completely fix the issue. We know that the downstream effects do have some impact on the pressure, and we can't change where the pressure ports are. We are going to try to fix the internal pressure changes.
We can't go off of this past launch as a reference for how much altitude/pressure will be off by, because we are now making changes that will reduce the delta. It would be bad to just blindly assume that the fixes will be "good enough". We won't be able to wind tunnel test before launch, and even if we could, the wind tunnel only goes to like 40 m/s.
I will check today with pelicanator data, but I think the altitude from integration will be off by 5-8 meters. It will also be an undershoot, which is best case scenario:
Extend airbrakes -> undershoot  altitude slightly -> retract airbrakes -> our altitude reading now goes back to pressure altitude -> it is slightly above target -> extend again

harshil21:
Superseded by #173

PR #164, opened 2025-02-24
Add pelicanator launch and fix mock

### LINKED ISSUE #162: Get mocks working on laptops
I can't run a mock on my laptop anymore, even after making sure pi libraries aren't in my pyproject.toml.

I think we could individually fix py files so they don't have these errors, but I would rather just only have files be imported in `main.py` as they are needed, e.g. so only import servo if we actually use it, so not during a mock.

harshil21:
Closes #162 

Fixing the mock made me realize there is another issue with the way we process IMU packets. An issue is coming up on that.

To do before merge:
- [x] Overhaul integration tests.
- [x] @wlsanderson can you confirm the pressure and temperature? The temperature from historical data seems wildly off.

### DIFF
--- a/launch_data/README.md
+++ b/launch_data/README.md
@@ -13,9 +13,8 @@ The timestamps of the LandedState packets were synced with the timestamps of the
 5. `legacy_launch_1.csv`: This was the first flight with an encoder and camera. The encoder clearly indicated that airbrakes deployed during coast, but the camera footage was indicated that they deployed during free fall. This launch told us that there is a massive latency coming directly from
 the IMU.
 6. `legacy_launch_2.csv`: We changed the IMU packet frequency to 50Hz for estimated packets, and 100Hz for raw packets. Unfortunately, this flight was a dud since our velocity was negative even in StandbyState, which indicated that the IMU had severe issues.
-7. `pelicanator_launch_1.csv`: This was the first flight of the Pelicanator rocket. The airbrakes deployed during coast, for roughly 4 seconds, and they retracted, successfully lowering our apogee by about 60 meters. The recovery system of the rocket failed however, thus the rocket crash landed and our Raspberry Pi 5 switched off, thus we lost about 22 seconds of data. The camera was in the payload which didn't work, so there's no footage of the airbrakes.
+7. `pelicanator_launch_1.csv`: This was the first flight of the Pelicanator rocket. The airbrakes deployed during coast for roughly 4 seconds, and then retracted, successfully lowering our apogee by about 60 meters. However, the recovery system of the rocket failed, causing the rocket to crash land. Our Raspberry Pi 5 switched off, resulting in the loss of about 22 seconds of data. The camera was in the payload, which didn't work, so there's no footage of the airbrakes.
 
-## Videos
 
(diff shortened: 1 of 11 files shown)


wlsanderson:
I fixed temp and pressure. Wasn't too far off. I never recorded the wind speed, but 12.81 kmh sounds about right... Might have been a little less but idk

tests/auxil/launch_cases.py
```diff
@@ -0,0 +1,294 @@
...
+    max_velocity: float | None = None
+    extensions: list[ServoExtension] = []
+    min_altitude: float | None = None
+    max_altitude: float | None = None
+    min_avg_vertical_acceleration: float | None = None
+    max_avg_vertical_acceleration: float | None = None
+    apogee_prediction: list[float] = []
+
+
+class LaunchCase:
```
JacksonElia, line 31: Nice this is clean af

tests/test_integration.py
```diff
@@ -62,6 +52,19 @@ def test_update(
...
+        elif launch_name == "legacy_launch_1":
+            launch_case = LegacyLaunchCase
+        elif launch_name == "genesis_launch_2":
+            launch_case = GenesisLaunchCase
+        elif launch_name == "interest_launch":
+            launch_case = InterestLaunchCase
+        elif launch_name == "pelicanator_launch_1":
+            launch_case = PelicanatorLaunchCase
+        else:
+            raise ValueError(f"Unknown launch name: {launch_name}")
```
JacksonElia, line 66: okay u can use a match statement here

JacksonElia (approved):
Good work with the integration tests.

tests/test_integration.py
harshil21: tbf since it's slower I don't see any reason to use it lol

launch_data/README.md
```diff
@@ -13,8 +13,8 @@ The timestamps of the LandedState packets were synced with the timestamps of the
 5. `legacy_launch_1.csv`: This was the first flight with an encoder and camera. The encoder clearly indicated that airbrakes deployed during coast, but the camera footage was indicated that they deployed during free fall. This launch told us that there is a massive latency coming directly from
 the IMU.
 6. `legacy_launch_2.csv`: We changed the IMU packet frequency to 50Hz for estimated packets, and 100Hz for raw packets. Unfortunately, this flight was a dud since our velocity was negative even in StandbyState, which indicated that the IMU had severe issues.
+7. `pelicanator_launch_1.csv`: This was the first flight of the Pelicanator rocket. The airbrakes deployed during coast for roughly 4 seconds, and then retracted, successfully lowering our apogee by about 60 meters. However, the recovery system of the rocket failed, causing the rocket to crash land. Our Raspberry Pi 5 switched off, resulting in the loss of about 22 seconds of data. The camera was in the payload, which didn't work, so there's no footage of the airbrakes.
```
harshil21, line 17:
```suggestion

## Videos
```
dumb copilot removed it

tests/test_integration.py
JacksonElia: its a test and its more readable + this thing is called like 5 times so performance is not a factor

PR #166, opened 2025-03-01
`MockIMU` upgrades, `--mock-servo` & `--mock-camera` for `uv run real`

### LINKED ISSUE #159: Fix jank 5am code
Since things were exclusively coded on the Pi last night, many (minor) things were not fully thought out and results in the mock not working on the pi or locally.

Things to fix:

- [ ] Dependencies: `lgpio` can only be successfully installed on the pi, couldn't get it working locally (kinda due to time constraints)
- [ ] Mock replay completely fails on pelicantor flight (!!)
- [ ] The new `-i` argument must be mutually exclusive with `-f`.
- [ ] `test_main.py` effectively fails on both arm64 and x86_64 in different ways (related to point 1). So `pytest_ignore_collect` needs to be deleted.

harshil21:
Renames `-r` to `-s`. This is more consistent with `-c`, which is you know, c for camera. So s should be for servo.

Replaces `-i` of `uv run mock` and instead adds `-s` and `-c` to `uv run real`. This helps with the confusion and makes argument exclusivity much more easier to handle. 

Also cherry picks upgrades from #106 since that is still far from done.

Merge after #164 

Closes #159


JacksonElia:
>Replaces -i of uv run mock and instead adds -s and -c to uv run real. This helps with the confusion and makes argument exclusivity much more easier to handle.

I don't know if I like this change. I would rather have uv run real always run with all hardware, and if we want to test anything individually, we do it via the mock.

harshil21:
> I don't know if I like this change. I would rather have uv run real always run with all hardware, and if we want to test anything individually, we do it via the mock.

the problem is that if the servo isn't attached, then the code will fail to communicate on the gpio pin and error out. 

`uv run mock -i` never made sense because it is impossible to handle the exclusivity with `-p` and `-f`. Running a mock should always mean using a MockIMU, and that's the main difference between `uv run real` and `uv run mock`.

wlsanderson (approved):

JacksonElia (approved):

harshil21:
I think I didn't update the readme for replacing -r with -s, can do in another PR

PR #167, opened 2025-03-02
Decent changes

### LINKED ISSUE #160: Make the logger flush more frequently
Due to the hard impact we had with the ground, our pi lost power. This caused us to lose a good portion of data due to our loggers internal buffering. Because the logger is on a different process, and we have a pi 5, we can afford to log more often.

JacksonElia:
Just doing some more tidying up

Closes #160


scripts/run_buzzer.py
harshil21, line 1: we kinda need this in the future maybe

JacksonElia: we still have the gui one, tho. Do u want both? Also, can I delete plots/?

harshil21: there is no gui for the buzzer?

airbrakes/constants.py
```diff
@@ -105,6 +105,38 @@ class DisplayEndingType(StrEnum):
...
+The signal to stop the IMU, Logger, and ApogeePredictor process, this will be put in the queue
+to stop the processes.
+"""
+
+# Formula for converting number of packets to seconds and vice versa:
+# If N = total number of packets, T = total time in seconds:
+# f = EstimatedDataPacket.frequency + RawDataPacket.frequency = 500 + 500 = 1000 Hz
+# T = N/f => T = N/1000
+
+IDLE_LOG_CAPACITY = 5000  # Using the formula above, this is 3.33 seconds of data
```
wlsanderson, line 128: should be 5.00 seconds of data now, not 3.33

airbrakes/context.py
```diff
@@ -58,30 +58,30 @@ def __init__(
         imu: BaseIMU,
         camera: Camera,
         logger: Logger,
-        data_processor: IMUDataProcessor,
+        data_processor: DataProcessor,
         apogee_predictor: ApogeePredictor,
     ) -> None:
         """
         Initializes AirbrakesContext with the specified hardware objects, Logger, IMUDataProcessor,
```
wlsanderson, line 65: if we are changing back to DataProcessor, just do a global search on "IMUDataProcessor", "IMU Data Processor", "IMU DataProcessor", etc.

airbrakes/main.py
```diff
@@ -26,33 +26,33 @@
...
 from airbrakes.telemetry.apogee_predictor import ApogeePredictor
-from airbrakes.telemetry.data_processor import IMUDataProcessor
+from airbrakes.telemetry.data_processor import DataProcessor
 from airbrakes.telemetry.logger import Logger
 from airbrakes.utils import arg_parser
 
 
 def run_real_flight() -> None:
     """Entry point for the application to run the real flight. Entered when run with
-    `uv run real` or `uvx --from git+... real`."""
```
wlsanderson, line 36: I think we should keep the backticks imo.

airbrakes/context.py
harshil21: oh damn was it data processor initially?

PR #168, opened 2025-03-03
Imu optimization

JacksonElia:
When making either a RawDataPacket or EstimatedDataPacket we were unnecessarily checking all of the fields for both of them. Instead, I made it so it only checks for raw fields when populating a RawDataPacket, and estimated fields when populating an EstimatedDataPacket.

Additionally, I thought of a further optimization that I might try if @harshil21 thinks it would actually help: Once we populate a field, we know for that data packet, we won't have to populate it again. So what if we made a list of fields that we check to add to a data packet, and as we add them to said data packet, we remove them from the list so that the fields at the end will have to do fewer comparisons?

### DIFF
--- a/airbrakes/telemetry/packets/imu_data_packet.py
+++ b/airbrakes/telemetry/packets/imu_data_packet.py
@@ -48,11 +48,11 @@ class EstimatedDataPacket(IMUDataPacket):
     It contains a timestamp and the estimated values of the relevant data points.
     """
 
+    estPressureAlt: float | None = None
     estOrientQuaternionW: float | None = None
     estOrientQuaternionX: float | None = None
     estOrientQuaternionY: float | None = None
     estOrientQuaternionZ: float | None = None
-    estPressureAlt: float | None = None
     estAttitudeUncertQuaternionW: float | None = None
     estAttitudeUncertQuaternionX: float | None = None
     estAttitudeUncertQuaternionY: float | None = None
--- a/airbrakes/hardware/imu.py
+++ b/airbrakes/hardware/imu.py
@@ -100,7 +100,7 @@ class IMU(BaseIMU):
         # - using 2 inner loops depending on the type of packet, this reduces the number of `if`
         # checks the interpreter has to do
         # - Using integers for comparison instead of strings (O(1) vs O(n) complexity)
-        # - Checking for raw data packets first, since that is 2x the frequency of estimated packets
+        # - Checking for raw data packets or estimated data packets, not both
         # - Using msgspec to serialize and deserialize the packets, which is faster than pickle
         # - High priority for the process
         while self.is_running:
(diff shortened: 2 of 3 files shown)


harshil21:
> we were unnecessarily checking all of the fields for both of them. Instead, I made it so it only checks for raw fields when populating a RawDataPacket, and estimated fields when populating an EstimatedDataPacket.

it was initially like this and you reverted it. What changed? 

>  So what if we made a list of fields that we check to add to a data packet, and as we add them to said data packet, we remove them from the list so that the fields at the end will have to do fewer comparisons?

how do you want to check from that list? a `getattr` call? That would be more expensive than just iterating through the conditions.

harshil21:
I also thought of using a mapping with the keys as the field names, and then the values as function to call which would set the attribute. This would be constant time. However I think the overhead of calling a function would offset that unfortunately.

airbrakes/telemetry/packets/imu_data_packet.py
```diff
@@ -48,11 +48,11 @@ class EstimatedDataPacket(IMUDataPacket):
     It contains a timestamp and the estimated values of the relevant data points.
     """
 
+    estPressureAlt: float | None = None
     estOrientQuaternionW: float | None = None
     estOrientQuaternionX: float | None = None
     estOrientQuaternionY: float | None = None
     estOrientQuaternionZ: float | None = None
-    estPressureAlt: float | None = None
     estAttitudeUncertQuaternionW: float | None = None
```
harshil21, line 56: how does this affect performance

JacksonElia:
>it was initially like this and you reverted it. What changed?

I never reverted it? I remember telling you and craig after doing these optimizations that you should do this

>how do you want to check from that list? a getattr call? That would be more expensive than just iterating through the conditions.

Yeah I'm not really sure. ~Maybe we could do something with a dict but it might end up not being as fast.~ Okay after thinking it through more, you're right. The only thing I can think of is have booleans for each field and once we check it, we make those booleans false, so our if statements look like `if field_bool and field_name == field`

~This might be faster because comparing strings are kinda expensive~ Nvm we are comparing ints, so I guess its fine

airbrakes/telemetry/packets/imu_data_packet.py
JacksonElia: it doesn't, I just thought it made more sense to have our more important field higher up like that, rather than sandwiched between quaternions.

harshil21:
> I never reverted it? I remember telling you and craig after doing these optimizations that you should do this

https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/146/commits/1f2e423db0d03e19661c4e8f9d286f2b008dfd78

harshil21:
I can merge after physically testing it on the pi.

JacksonElia:
Lol I guess I accidentally did when switching over to the constants, woops

JacksonElia:
Nah lets just merge it now, we can test everything on decent-changes before merging

PR #172, opened 2025-03-05
Use `os.fsync()` to guarantee disk write

### LINKED ISSUE #171: Investigate more on why data was cutoff for Pelicanator flight
See my comment in https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/pull/167/commits/7aedc75e10aab39414bc930d5c8a6ced16aba7ac

Essentially what's happening is:

1. Python already flushes to the file more frequently than we call `flush()`.
2. `flush()` only guarantees that Python hands over the data to the Linux kernel. It does not guarantee that contents have been written to the file yet.
3. The data now sits in kernel’s page cache (i.e. RAM). Depending on system load, the filesystem (I believe it's ext4 on the pi), and the filesystem mount options (e.g. `data=ordered`) the OS decides when it's a good time to commit the data.

Proposed fix:

1. We should use `os.fsync()` to force the OS to write to the file, instead of this flush() which evidently does not help at all.
2. Test more on the pi 5 before/after this fix and confirm that this will actually fix it.
3. Design a test to SIGKILL the process and then check the file contents.

harshil21:
Closes #171 

Unfortunately there is no easy way to test that this works without changing kernel parameters at runtime like `vm.dirty_writeback_centisecs` to a very high value temporarily. 

Even then, reading a file to check if stuff was written will not work since the file will be read directly from the dirty pages. One way around that would be to specify the [`O_DIRECT`](https://manpages.debian.org/bookworm/manpages-dev/open.2.en.html#O_DIRECT) flag when reading a file. However this needs to be aligned with the block size of the filesystem (usually 4096 bytes, you can check this by running `fdisk -l`), and also the data returned will be in bytes.

There is also no way to specifically check the dirty page size for a file directly.

So while it should be possible to write a test for this, I'm not doing it since I don't want to mess with my kernel parameters every time I run the test suite.

### DIFF
--- a/airbrakes/constants.py
+++ b/airbrakes/constants.py
@@ -113,8 +113,9 @@ LOGS_PATH = Path("logs")
 """The path of the folder to hold the log files in."""
 TEST_LOGS_PATH = Path("test_logs")
 """The path of the folder to hold the test log files in."""
-NUMBER_OF_LINES_TO_LOG_BEFORE_FLUSHING = 100
-"""The number of lines we log before manually flushing the buffer."""
+NUMBER_OF_LINES_TO_LOG_BEFORE_FLUSHING = 1000  # 1 second of data
+"""The number of lines we log before manually flushing the buffer and forcing the OS to write
+to the file."""
 
 STOP_SIGNAL = "STOP"
 """
--- a/airbrakes/telemetry/logger.py
+++ b/airbrakes/telemetry/logger.py
@@ -2,6 +2,7 @@
 
 import csv
 import multiprocessing
+import os
 import signal
 from collections import deque
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


airbrakes/constants.py
```diff
@@ -113,8 +113,9 @@ class DisplayEndingType(StrEnum):
 """The path of the folder to hold the log files in."""
 TEST_LOGS_PATH = Path("test_logs")
 """The path of the folder to hold the test log files in."""
-NUMBER_OF_LINES_TO_LOG_BEFORE_FLUSHING = 100
-"""The number of lines we log before manually flushing the buffer."""
+NUMBER_OF_LINES_TO_LOG_BEFORE_FLUSHING = 1000  # 1 second of data
```
JacksonElia, line 116: it should be lower than this, 100 seems good to me

harshil21:
Tested on the pi before/after this change and confirm that it works. 

Before this change it definitely wrote to the file after ~20 seconds, confirming our suspicion that the pi switched off after landing hard.

wlsanderson (approved):

PR #179, opened 2025-03-16
Use JIT / LLVM 20 built Python for performance improvements

### LINKED ISSUE #178: Use JIT compiled Python for performance improvements
See https://github.com/astral-sh/uv/releases/tag/0.6.6.

I can run the benchmarks and check if there's a noticeable improvement. This should also help the IMU tight loop and we won't need to worry about optimizing that further. We should also enable this in the CI.

harshil21:
Closes #178 

Turns out most performance improvements comes from building Python with LLVM 20 instead of 19, which had performance regressions. Running a replay with `-f` on genesis_launch_2.csv took ~2.56s previously, and now it takes ~2.3s on my machine, with about the same results with/without `PYTHON_JIT=1`. 

Latest `uv` python's automatically have this - but only for Linux Intel x86_64 currently.. so performance improvements are not realized for the raspberry pi, yet..


harshil21:
The CI was failing because I had found a bug with the uv managed pythons. Reported that and it's now resolved. Tracking support for better arm64 python builds here - https://github.com/astral-sh/python-build-standalone/pull/484 ?

harshil21:
> bro there are 19 lines added how was this 17 commits

1 line = 1 commit for optimization you know

PR #184, opened 2025-03-29
Fix unix timestamp on pelicanator launch 1

wlsanderson:
Someone hit backspace on a data row in the csv before uploading it to launch data


harshil21 (approved):
oopsies 

wasn't caught earlier cause we don't do anything with that time

JacksonElia:
lol nice catch

PR #187, opened 2025-04-05
Changing Convergence Thresholds

wlsanderson:
loosening the convergence thresholds improves both the accuracy of prediction and the convergence time. These values are fine-tuned for pelicanator launch 1 and 2, but also significantly improve prediction and convergence time on genesis launches and interest launch.


harshil21 (approved):
we never fine tuned this value after genesis launches. The value was initially set based only on interest launch. So now it's much better for all flights!

PR #272, opened 2026-01-22
Add firm

JacksonElia:


airbrakes/context.py
```diff
@@ -76,15 +74,15 @@ def __init__(
             rocket will be based on the processed data.
         """
         self.servo: BaseServo = servo
-        self.imu: BaseIMU = imu
+        self.firm: FirmClient = firm
```
JacksonElia:
```suggestion
        self.firm: BaseFIRM = firm
```

airbrakes/context.py
```diff
@@ -144,11 +142,11 @@ def update(self) -> None:
         # *may* not be the most recent data. But we want continuous data for state, apogee,
         # and logging purposes, so we don't need to worry about that, as long as we're not too
         # behind on processing
-        self.imu_data_packets = self.imu.get_imu_data_packets()
+        self.firm_data_packets = self.firm.get_data_packets()
 
         # This should not happen generally, since we wait for IMU packets. Only happens at the end
```
JacksonElia:
```suggestion
        # This should not happen generally, since we wait for FIRM packets. Only happens at the end
```

JacksonElia (commented):
Good job so far! We can start working on actually testing/mocking FIRM soon.

airbrakes/hardware/firm.py
```diff
@@ -0,0 +1,53 @@
...
+from airbrakes.constants import FIRM_BAUD_RATE, FIRM_PORT, FIRM_SERIAL_TIMEOUT_SECONDS
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+class FIRM(BaseFIRM):
+    __slots__ = ("_log_file_path", "firm_client", "is_pretend")
+
+    def __init__(self, is_pretend=False, log_file_path: Path | None = None):
```
harshil21: having pretend in the real firm.py doesn't make sense imo, it should be in mock

airbrakes/main.py
```diff
@@ -178,30 +180,3 @@ def run_flight_loop(
...
-    #     -s, --real-servo   : Uses the real servo instead of a mock one.
-    #     -f, --fast-replay  : Runs the replay at full speed instead of real-time.
-    #     -p, --path <file>  : Specifies a flight data file to use (default is the first file).
-
-    # Global options for all modes:
-    #     -d, --debug   : Runs without a display, allowing inspection of print statements.
-    #     -v, --verbose : Enables a detailed display with more flight data.
-
-    args = arg_parser()
-    run_flight(args)
```
harshil21: we need some python options, e.g. `python --help`, so we need the legacy way of running.

airbrakes/utils.py
```diff
@@ -80,113 +80,74 @@ def deadband(input_value: float, threshold: float) -> float:
 def arg_parser() -> argparse.Namespace:
     """
     Handles the command line arguments for the main Airbrakes program.
-
-    :return: The parsed arguments as a class with attributes.
     """
-    # We require ONE and only one of the 3 positional arguments to be passed:
-    # - real: Run the real flight with all the real hardware.
-    # - mock: Run in mock replay mode with mock data.
```
harshil21: don't delete comments

PR #336, opened 2026-06-01
Make test less flakey

JacksonElia:
This apogee predictor test kept occasionally failing (e.g. https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/actions/runs/26597745805/job/78373085744 and https://github.com/NCSU-High-Powered-Rocketry-Club/AirbrakesV2/actions/runs/26704573602/job/78703288115?pr=334) I think these changes should make it fail barely ever unless it like really lags

I ran the tests 5 times and they passed every time

### DIFF
--- a/tests/test_components/test_apogee_predictor.py
+++ b/tests/test_components/test_apogee_predictor.py
@@ -90,8 +90,7 @@ class TestApogeePredictor:
         """
         Integration-ish test of the apogee predictor using the real HPRM
         call. These tests assume that HPRM works perfectly. To test the actual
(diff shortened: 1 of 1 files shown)


harshil21 (approved):
hard to believe that this would fix it since apogee prediction is ~instant and so is sending the data over. But I don't have the resources to look more into this right now, so lgtm.

JacksonElia:
The fix here is less increasing the timeout to 2.5 sec and more making it wait to get all of the prediction packets and choose the last one rather than just getting the first prediction it can get

PR #342, opened 2026-06-10
refactored servo, baseservo, mockservo classes

### LINKED ISSUE #335: Simplify servo code
Right now looking at the `BaseServo`, `MockServo`, and `Servo` classes, the logic is too complex for no real benefit. We should refactor `BaseServo` to behave like an interface with minimal to no logic/implementations like `BaseFirm`.

I would also like to remove the no buzz logic to make way for the extension lookup table mentioned in #332. 

Also, we are no longer using an encoder, so all of that code should be removed.

Here is a general outline of how I would go about it:
- Make BaseServo have the following method headers: extend_airbrakes(), retract_airbrakes(), get_battery_volts(), and get_system_current_milliamps()
- Delete the encoder_position  fields from the ServoDataPacket and LoggerDataPacket  --> in context.py and logger.py you'll need to remove the parts where you set this field in the data packets
- Make the Servo and MockServo classes implement the methods you declared in BaseServo.  For now, in MockServo you can just make the extend and retract methods pass, and then the getter methods return 0. In Servo  you can make the extend and retract methods just set the servo extension to the ServoExtension.MAX_EXTENSION and ServoExtension.MIN_EXTENSION  respectively. You will need to copy a lot of the pwm/duty cycle logic from the current BaseServo class

So basically the idea of this is that you're moving a lot of the implementation into the Servo class rather than splitting it across BaseServo and Servo. And you're also just simplifying it a lot by removing the encoder and no buzz stuff.

manujay3:
done for issue of simplifying servo code
closes #335


JacksonElia:
Nice work! I will look at this more later, but can you try to get the unit tests passing?

airbrakes/data_handling/logger.py
```diff
@@ -119,10 +119,10 @@ def _convert_unknown_type_to_str(obj_type: Any) -> str:
...
     @staticmethod
     def _prepare_logger_packets(
-        context_data_packet: ContextDataPacket,
-        servo_data_packet: ServoDataPacket,
-        firm_data_packets: list[FIRMDataPacket],
-        apogee_predictor_data_packet: ApogeePredictorDataPacket | None,
+        context_data_packet: 'ContextDataPacket',
+        servo_data_packet: 'ServoDataPacket',
+        firm_data_packets: list['FIRMDataPacket'],
+        apogee_predictor_data_packet: 'ApogeePredictorDataPacket' | None,
```
JacksonElia: Why did you wrap all the types in single quotes? I forget what it does

pranitlokineni: Pretty sure it's for future referencing. But I don't think it's necessary, since 2 classes aren't importing each other, right? Unless there is another use for it.

JacksonElia: yeah I think you're right. @manujay3 can you revert all the places where you added these?

airbrakes/context.py
```diff
@@ -77,18 +77,18 @@ def __init__(
             predicts what the apogee of the rocket will be based on the
             processed data.
         """
-        self.servo: BaseServo = servo
-        self.firm: BaseFIRM = firm
-        self.logger: Logger = logger
-        self.data_processor: DataProcessor = data_processor
-        self.apogee_predictor: ApogeePredictor = apogee_predictor
+        self.servo: 'BaseServo' = servo
```
pranitlokineni: Correct me if I'm wrong, but now that all the functionality is in the Servo class, servo should be initialized as an instance of Servo, not BaseServo.

JacksonElia: This should still be `BaseServo` because a `MockServo` could get passed in when running a mock

manujay3:
fixed the single quotes

airbrakes/context.py
```diff
@@ -53,11 +53,11 @@ class Context:
...
-        servo: BaseServo,
-        firm: BaseFIRM,
-        logger: Logger,
-        data_processor: DataProcessor,
-        apogee_predictor: ApogeePredictor,
+        servo: 'BaseServo',
+        firm: 'BaseFIRM',
+        logger: 'Logger',
+        data_processor: 'DataProcessor',
+        apogee_predictor: 'ApogeePredictor',
```
JacksonElia: you still have these quotes in other parts of the code, remove all of them please

JacksonElia: you can look at the `files changed` in the PR to see everything you've changed.

JacksonElia (commented):
Can you also try and get the unit tests working? Lmk if you have any questions with that.

tests/test_hardware/test_servo.py
```diff
@@ -19,132 +15,54 @@ class TestBaseServo:
...
 
     def test_start(self, servo):
         """
-        Tests that start() activates the PWM signal at the duty cycle
-        corresponding to MIN_NO_BUZZ, leaving current_extension unchanged.
+        Tests that start() executes safely.
         """
-        # Before start(), the duty cycle should be 0 (PWM not yet started)
-        assert servo.servo.duty_cycle == 0.0
         servo.start()
```
JacksonElia: these all got deleted cause they we're working, but we'll want to write good unit tests for the LewanLib pr once pranit makes that

JacksonElia (approved):

JacksonElia:
I think copilot's messages are mostly valid but @pranitlokineni you can just fix this in the lewan lib PR

pranitlokineni (approved):

PR #353, opened 2026-08-04
Make Renovate PR monthly and group all deps in that PR

harshil21:
Renovate will now upgrade all deps in one PR, once a month.


JacksonElia (approved):
hardshil might be the goat

PR #359, opened 2026-09-01
Handle transonic effects

### LINKED ISSUE #333: Figure out how to deal with transonic/supersonic speeds
In IREC it is possible our rocket exceeds mach and is guaranteed to hit at least transonic speeds. This means that our pressure sensor won't be reliable. To fix this, we should probably rely on integrating acceleration for both velocity and altitude until we are safely below transonic speeds. We might also not want to deploy airbrakes at transonic or higher speeds. I think we might want to add another state to the FSM to account for this rather than just putting even more logic in Coast.

JacksonElia:
So the main point of this PR is to make it so if the rocket is going at speeds close to or greater than where transonic effects start happening (~mach .7) we start integrating for altitude until we go slower again. But then I also did some other quality of life stuff like making it so there's now a field that gets logged for if we are integrating for altitude and also logging all of the dataprocessor fields.

Closes #333


airbrakes/data_handling/data_processor.py
```diff
@@ -212,42 +217,55 @@ def _calculate_current_altitudes(self) -> npt.NDArray[np.float64]:
...
-                    for data_packet in self._data_packets
-                ],
-            )
-        else:
-            # If we don't have an initial altitude, we just use the raw altitudes .-.
-            altitudes = np.array(
-                [data_packet.est_position_z_meters for data_packet in self._data_packets],
+        altitudes = np.empty(len(self._data_packets), dtype=np.float64)
+        previous_altitude = self._previous_altitude
+        self._integrating_for_altitudes = []
```
pranitlokineni: Why Initialize _integrating_for_altitudes with all "F" in the first update if it just gets reset here?

JacksonElia: thats cause all of the values in data_processor have default values for like 1 datapacket, so I think it's fine. Maybe later we can revisit this and see if we should change it for all of them tho

pranitlokineni (approved):

PR #360, opened 2026-09-05
attempting to implement deploymentlookuptable

### LINKED ISSUE #332: Make deployment lookup table
Our airbrakes have proven themselves to be very effective, capable of taking off more than 1000 feet. They can deploy and retract effectively for rocket speeds experienced in Student Launch, but most likely can't handle the max speeds present in IREC. To safeguard against deploying at a speed potentially higher than what they can effectively retract at, we should make a rocket speed vs extension lookup table for the airbrakes.

A simple approach to this would be kinda just doing our best guess about what is appropriate, e.g. [0 m/s: 100% extension, 150 m/s: 100% extension, 200 m/s: 75% extension, 250 m/s: 10% extension] and so on.

I think a better approach would first be making an extension-airbrake surface area lookup table (because the relationship between extension and surface area exposed isn't linear), and then doing a rudimentary flat-plate drag calculation and only letting the airbrakes deploy to the maximum force they can withstand. So if through testing we determine they can only withstand a total of 100 lbs, we only let them deploy to an extension that they would experience that or less.

Here is how I would go about it:

1. You're going to want to get familiar with BaseServo and Servo. I had Manu clean up some of our servo code, but that isn't merged yet so you're gonna want to look at the SimplifiedServo branch. Spend some time understanding what inheritance and abstract base classes are (pasting these files into chatgpt and asking questions is a good way to start)
2. Next, decide in what format you want your lookup table to be in, maybe a 2d array using lists, maybe using some libraries (I think numpy has functions that let you do linear interpolation with numpy arrays)
3. Make a placeholder lookup table for extension percentage vs surface area, just guess on what values seem reasonable (when fully extended, airbrakes have about 32 square inches outside of the rocket), and add it to constants.py
4. In BaseServo change the extend_airbrakes method to also have a parameter for the rocket's velocity
5. Add another method to BaseServo for calculating the deployment extension based on the rocket velocity, you will actually implement this method (implement in coding means to write the code for it, so for example all of the other methods in BaseServo don't have implementations). Make this method private (ask chatgpt what private vs public means in coding and specifically in python) and have it take in a velocity and return an extension between 0 and 1. This method is what is using your lookup table and doing linear interpolation.
6. You will also have to make another method to convert the extension percentage into degrees as when we extend airbrakes, we are just setting the angle of it to what we measured was extended (remember when we were getting the servo angles). This method can also be private and you could put it in BaseServo  or Servo depending on how you wrote your code
7. Update the extend_airbrakes method in Servo to use these new methods you made (a method is just a function that is in a class, so you could also just call them functions im just using slang)

manujay3:


airbrakes/hardware/servo.py
```diff
@@ -84,9 +83,10 @@ def stop(self) -> None:
         # Release the gpio pin back to the kernel
         self._servo_line.release()
 
-    def extend_airbrakes(self) -> None:
-        """Command the physical servo to its maximum extension."""
-        self._servo.move_time_write(SERVO_MAX_EXTENSION, 0)
+    def extend_airbrakes(self, velocity: float) -> None:
```
yarielhdez771-cmd: sohuld this not have the same argument (velocity_meters_per_s) as the base servo class?

yarielhdez771-cmd (commented):
yeah looks fine to me, just got the question about the arguments in mock/servo classes

airbrakes/hardware/servo.py
JacksonElia: good catch, yea

pranitlokineni (approved):
Looks good. Is it worth accounting for horizontal velocity also? If the rocket ends up having a ton of tilt when launching, the magnitude could be pretty different than just the vertical component.

JacksonElia:
@pranitlokineni thats good to keep in mind yeah, but we don't really have horizontal vel accurately rn, we should get it later and use it. We also should use it for apogee prediction.

